PR #28, opened 2014-12-19
remove the other 1 word alias lines

graysky2:
In reference to the following comment and advise therein: https://github.com/grml/grml-etc-core/commit/e020ae7784364f9a29b01dfc785b503f1adc0e87


mika:
Thanks @moviuro, I basically agree with the suggestion to remove the 'l' alias as well, but this one is quite commonly used AFAIK, so I tend to keep it as is.

PR #30, opened 2015-01-06
zshrc: new abbreviation expansion

moviuro:
NE expands to a grep expression that filters empty lines and lines starting with a # (stands for Not Empty)
Useful when cat-ing conf files (among others)


moviuro:
@ft I think you can close this pull request

PR #31, opened 2015-02-07
moviuro's pull requests

moviuro:
We use a consistent syntax and don't fire up a new process.


mika:
Updated PR in https://github.com/grml/grml-etc-core/pull/34, closing this therefore.

moviuro:
errr, there were some more commits in here

PR #32, opened 2015-04-09
always use classify for ls output

eworm-de:


moviuro:
wouldn't it work to define : `alias _ls=command ls -F` and then `alias ll=_ls -l` and finally, `unalias _ls` ?
(I'm not even sure it works, just shooting in the dark, here)

mika:
Thanks for your PR! I'm pretty sure though that soonish someone will come and wants to get that change reverted if we'd apply it, because those aliases used to be that way as they are since quite some time already. Changes to such "central" aliases are pretty tough to get right for everyone, so I'm afraid we can't merge this PR. If you want to use this configuration place it in your personal .zshrc.local config. (To be honest, I'm actually even inclined to drop the '-F' from the ls alias.)

eworm-de:
I would be happy to drop '-F' from aliases as well. But IMHO it should be consistent in one way or another. I expext 'ls -l' and 'll' to look the same, but currently it does not.

moviuro:
We could use the `${ls_options:+"${ls_options[*]}"}` part to add `-F` in the `.zshrc.pre`, I'll work into it.

moviuro:
See https://github.com/moviuro/grml-etc-core/tree/ls-classify & c05661e1d944f920bf3de53f2ba83e4265ddc32d

PR #33, opened 2015-04-09
add issolaris()

ttyS4:
I am working in an environment where I could make use of an 'issolaris' function.
Adding it here similar to the others.


mika:
Thanks for the PR, got merged!

PR #34, opened 2015-04-12
Unify and simplify assignment of options to ls and grep

moviuro:


mika:
Thanks, I've cherry-picked everything except for the the "simplify simple test functions" and "consistent look of isXYZ functions' because ft and me think that it's better to have this explicit than implicit.

Thanks for all your patches!

PR #37, opened 2015-04-13
zshrc: we make sure that aliases refer to 'command ls'

moviuro:


moviuro:
Stoopid me, too used to commit-ing on zshrc that I wrote the commit wrong!

PR #40, opened 2015-04-19
zshrc: fix alias conflict

trrbl:
ag command is used by a popular searching tool https://github.com/ggreer/the_silver_searcher

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -2578,9 +2578,9 @@ if [[ -r /etc/debian_version ]] ; then
     #a3# Execute \kbd{aptitude install}
     salias ati="aptitude install"
     #a3# Execute \kbd{apt-get upgrade}
-    salias ag="apt-get upgrade"
+    salias agg="apt-get upgrade"
     #a3# Execute \kbd{apt-get update}
-    salias au="apt-get update"
+    salias agd="apt-get update"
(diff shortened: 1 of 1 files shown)


mika:
Uhm, this is a tricky one.The 'au' and 'ag' aliases are very well established among Debian folks using grml-zshrc. :-/ What we *could* do is check whether au and ag exist in $PATH and if so then don't set up the aliases. Would this work for you? If so then I'd discuss this amongst the Grml team.

trrbl:
Check aliases in $PATH is interesting way to solve the problem. Now I use unalias right after loading grml-zshrc. Anyway, it would be nice if the problem will be solve by any acceptable way :) Thanks

mika:
FTR, we (Grml team) decided to not accept the PR as is (because of the way it modifies behaviour for longtime users of Grml/grml-zshrc). We also won't modify the aliases depending on what's inside $PATH to not force incosistent behaviour. Instead what we think would make sense is create a new option/env variable check for grml-zshrc which doesn't add such short aliases then. Leaving the PR open for now until someones finds time to implement this change.

moviuro:
check : https://github.com/grml/grml-etc-core/pull/42

mika:
I think https://github.com/grml/grml-etc-core/pull/59 fixes this issue, sorry for taking so long to find a working solution for this issue. Closing this issue now, please reopen if you're aware of anything further, thanks everyone!

PR #42, opened 2015-04-22
zshrc: add `ag' alias only if `ag' is not in $PATH

moviuro:
Add agg instead if otherwise


moviuro:
deemed unacceptable on IRC, still the issue remains.

ft:
Like the other functions and aliases, "ag" is documented in the setups manual page. So we're
not sneaking up on anyone with its use. Our use of the name also precedes the lifetime of the
external application — which, although being useful, is no standard program at all.

Since "ag" is an alias, you can call ag like this even with the alias in place:

- =ag
- \ag
- command ag

And if you absolutely want to change the name of the alias, use our .zshrc.local file:

unlias ag
alias agg='apt-get upgrade'

So, for the record: This is not an issue as far as we are concerned.

PR #44, opened 2015-04-22
zshrc: check SSL certificates in `simple-extract'

moviuro:
Espescially because it's an _online_ archive...


mika:
Sorry for missing this one, I fixed this in commit 43d45335db2dcfcb8c143a248057636508bfc4ae without seeing your PR, sorry for that :(

PR #47, opened 2015-05-18
fix ssh remote host name completion with more than one pattern

eworm-de:
Host declaration may contain more than one pattern given after the
keyword, separated by whitespaces. For correct completion split the
string.


mika:
Nice catch, thanks for the PR! :+1:

PR #48, opened 2015-06-01
Avoid placing empty components in PATH; reorder root's PATH

akorn:
zsh treats an empty PATH component ("::") as ".", which creates a security hole.

Removing the unnecessary quotes around "$ADDONS" and "$path[@]" avoids putting an empty array element into the path array. The substitutions are still whitespace safe because zsh doesn't split words unless told to do so explicitly.

As agreed on #grml, root's PATH should also list /usr/local/sbin and bin before /sbin, /bin, /usr/sbin and /usr/bin; and there is no harm in adding $HOME/bin unconditionally as the first PATH component.

### DIFF
--- a/etc/zsh/zshenv
+++ b/etc/zsh/zshenv
@@ -85,19 +85,20 @@ if (( EUID != 0 )); then
     /sbin
     /usr/local/games
     /usr/games
-    "${ADDONS}"
-    "${path[@]}"
+    ${ADDONS}
+    ${path[@]}
   )
 else
   path=(
+    $HOME/bin
+    /usr/local/sbin
+    /usr/local/bin
     /sbin
     /bin
     /usr/sbin
     /usr/bin
-    /usr/local/sbin
-    /usr/local/bin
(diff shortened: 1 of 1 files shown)


akorn:
I suppose that needs to be done anyway, just to be sure, because one of the expanded variables may (at some point in the future) be an array that includes empty elements.

Still, the quotes around ```${ADDONS}``` are superfluous, so why not remove them?

ft:
I'd like to keep it like that, because the semantics are clear if the expansion
is used double-quoted. You don't have to figure out if "shwordsplit" is set or
not.

It may be superfluous (it is), but it doesn't hurt if we weed out empty $path
entries later anyway.

akorn:
I don't feel strongly about the quotes; I'm fine with just your proposed sanitization line.

However, let's keep the part where /usr/local has precedence over / and /usr.

ft:
Sounds good.

mika:
Fixed via https://github.com/grml/grml-etc-core/issues/49

PR #49, opened 2015-06-02
Akorn path sanitization 1

akorn:
In #48 @ft suggested that the quotes in the expansions of `ADDONS` and `path` be re-added and that `path` be sanitized using a substitution that drops empty components.

This newer patch does it that way.


mika:
Thanks for the PR! I've just squashed the commits and slightly adjusted the commit message.

PR #53, opened 2015-09-22
zshrc: fix screen alias

dmitrivereshchagin:
- Using `${commands[screen]}` which is expanded in double quotes makes it impossible to use custom screen installation or wrapper.  Use `screen` instead.
- Screen reads configuration from `$HOME/.screenrc` by default so it is redundant to set alias for this.


mika:
Sorry for the delay, we also received your initial patch to the grml-etc-core mail address and even discussed this on IRC and it was ack-ed but before managing to merge it I got ill, now finally being at the computer again I took care of rebasing and merging it. :) Thanks for your work and PR! :+1:

PR #54, opened 2015-11-03
tmux: remove obsolete mouse-select-pane option

sometimesfood:
This pull request removes the obsolete `mouse-select-pane` option from the tmux config.

Since the `mouse-select-pane` is deactivated in the config anyway and is off by default, this does not affect the current behaviour in any way. It does, however, fix the `unknown option: mouse-select-pane` warnings produced by tmux 2.1.


mika:
Nice, thanks! :)

PR #55, opened 2016-01-14
Adding a tiny bit to aliases

moviuro:
Comment the obscure options ;
Add `ls -v` where available (natural sorting)

``` sh
$ touch 1 2 10; ls
1 10  2
$ ls -v
1  2 10
```

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -441,7 +441,8 @@ elif ls -G / >/dev/null 2>&1; then
 fi
 
 # Natural sorting order on GNU ls(1)
-if ls -v / >/dev/null 2>&1; then
+# OSX and IllumOS have a -v option that is not natural sorting
+if islinux && ls -v / >/dev/null 2>&1; then
     ls_options+=( -v )
 fi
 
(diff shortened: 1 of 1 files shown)


xaocon:
I'd be a little worried that ls -v / might catch on some versions of ls that don't support it as natural sorting but I don't know enough about the history or different implementations to say it would be a problem.

moviuro:
This is exactly why I use this line: https://github.com/moviuro/grml-etc-core/commit/b78f11c674d58383a4be2188f53a9f6a5d050320#diff-adada33f082dd4fb1e2b5a3b99b183e0R441 . It won't print anything (ever) and if it fails, nothing further happens. We know it works because we already do it with colors (`-G` an `--color=auto`).

KellerFuchs:
Not sure `islinux()` is the right thing to check for: it could still be non-GNU coreutils ...

moviuro:
Now should be better. No issues to report on FreeBSD nor Archlinux.

KellerFuchs:
Looks good to me, though I am but a random fox on the Internet.

KellerFuchs:
@mika Looks like this has been sitting here for a year.
Could you have a look and comment or merge?

mika:
Sorry that this took so long, I like the change and just merged it!
Thanks a ton for the PR and your patience :)

PR #56, opened 2016-02-24
Lookup: Fix warn_create_global warnings

nbraud:
This fixes warnings that occur whenever opening a login shell on Debian Stretch.


nbraud:
Nevermind, this is fixed in `master` and only occurs in the current `grml-stable` repo.

PR #57, opened 2016-02-24
Make ambiguous aliases optional

nbraud:
Having a stab at https://github.com/grml/grml-etc-core/pull/40#issuecomment-95341878
Also fixes a minor issue with `mdstat`

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -2543,7 +2543,10 @@ else
     alias l='command ls -l'
 fi
 
-alias mdstat='cat /proc/mdstat'
+if [[ -r /proc/mdstat ]]; then
+    alias mdstat='cat /proc/mdstat'
+fi
+
 alias ...='cd ../../'
 
 # generate alias named "$KERNELVERSION-reboot" so you can use boot with kexec:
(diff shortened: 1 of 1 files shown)


nbraud:
I chose on purpose not to consider the “listing stuff” commands as ambiguous, given how widespread they are (even in non-ZSH shells).

mika:
Hm, I'm not yet sure whether GRML_NO_AMBIGUOUS_ALIASES is probably the best possible name, though nothing really better came to my mind yet. Also if we add it to the zshrc we have to document it in the manpage accordingly.

mreinhardt:
👍 for this change, I use https://github.com/ggreer/the_silver_searcher and so unset these in my local.

An alternative way to mitigate them would be to remove them from /etc/zsh/zshrc and put them in https://github.com/grml/grml-etc-core/blob/master/etc/skel/.zshrc commented out instead.

mreinhardt:
I too always unset these in my `.zshrc.local` and would like to see this merged.

Other potential env var names:

```
GRML_NO_DEBIAN_ALIASES
GRML_NO_APT_ALIASES
GRML_UNSET_APT_ALIASES
```

Also @nbraud here's where to update the manpage for this pull request: https://github.com/grml/grml-etc-core/blob/master/doc/grmlzshrc.t2t#L43-L78

nbraud:
@mreinhardt Thanks, this completely fell off my radar  :V

nbraud:
Seems I removed my ability to update this PR by deleting my “fork” of this repo.

PR #58, opened 2016-05-04
zshrc: dirstack: skip for ZSH_SUBSHELL

blueyed:
Otherwise `(cd foo)` would add `foo` to the stack, which is probably not
what you would expect.

Side note: the hook should be installed through `add-zsh-hook` probably?!

I am not using grml myself, but had taken this part a while ago.

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -1613,6 +1613,7 @@ if zstyle -T ':grml:chpwd:dirstack' enable; then
     }
 
     chpwd() {
+        (( ZSH_SUBSHELL )) && return
         (( $DIRSTACKSIZE <= 0 )) && return
         [[ -z $DIRSTACKFILE ]] && return
         grml_dirstack_filter $PWD && return


ft:
Yes, good catch. This is clearly supposed to be an interactive-only feature, so
this is the right thing to do. I've merged this upstream and will close the request
therefore.

About add-zsh-hook: "add-zsh-hook" was added in zsh version 4.3.4. And we
kind of try to support zsh versions 3.1.7 and up and only use newer features
with version-tests around them. And in earlier versions of the setup we even
tried to support way older versions. So that's why the dirstack stuff was imple-
mented directly in chpwd when it was first introduced.

We do use add-zsh-hook in a few places, though. And since it is obviously
the cleaner solution we may well switch to it at one point. I'd kind of like to
lift the version requirement of our setup to, say, 4.3.7 (which is the version
a major bug in add-zsh-hook was fixed) or maybe to 4.3.10 (which is the
version in Debian old-old-stable) or maybe even 4.3.17 (which is the version
in Debian old-stable). I guess most other OSs will have a version at least
that recent, although you never know about centos…

If we do that, we can clean up a few other places as well.

blueyed:
Thanks for merging and the explanation.

I got the impression (from/via 4ac3e89 where the `is42` check was removed) that support for older versions was dropped by now (but had not investigated, only seen that it was removed/changed).

ft:
That particular change removes a workaround for pre-4.2 shells by changing
the implementation in a way that it is not required anymore. And since that
was the only place the `is42()` was used, it was also removed.

That being said, I used to have a collection of builds of all zsh versions ever
released and checked compatibility regularly. I don't maintain that collection
anymore and thus don't really check backwards compatibility a lot anymore.

That is another reason, we should probably increase the minimum version
requirement by quite a bit.

PR #59, opened 2016-11-03
Make ambiguous aliases optional

nbraud:
This is a rehash of #57


mika:
Thanks for finalizing that, also thanks to @mreinhardt for feedback and @ft for reviewing!

PR #60, opened 2017-01-09
etc/zsh/zshrc: date invocation in bk() now uses local timezone instead of UTC

moviuro:
This change works across all my OSes (GNU/Linux, FreeBSD and OpenBSD) and should even be POSIX.

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -3203,7 +3203,7 @@ fi
 #f5# Backup \kbd{file_or_folder {\rm to} file_or_folder\_timestamp}
 bk() {
     emulate -L zsh
-    local current_date=$(date -u "+%Y-%m-%dT%H:%M:%SZ")
+    local current_date=$(date +"%Y-%m-%dT%H:%M:%S%z")
     local clean keep move verbose result all to_bk
     setopt extended_glob
     keep=1


etc/zsh/zshrc
```diff
@@ -3203,7 +3203,7 @@ fi
 #f5# Backup \kbd{file_or_folder {\rm to} file_or_folder\_timestamp}
 bk() {
     emulate -L zsh
-    local current_date=$(date -u "+%Y-%m-%dT%H:%M:%SZ")
+    local current_date=$(date +"%Y-%m-%dT%H:%M:%S%z")
```
KellerFuchs: `%z` doesn't seem to be in [POSIX.1-2008](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/date.html).

What's the issue with using UTC?

mika: What about using:

```
date +"%Y-%m-%dT%H:%M:%S%Z"
```

instead, where "%Z" is in POSIX?

jkirk: JFTR:
```
jkirk@invincible ~ % date -u "+%Y-%m-%dT%H:%M:%SZ"
2018-05-24T16:25:13Z
jkirk@invincible ~ % date +"%Y-%m-%dT%H:%M:%S%z"
2018-05-24T18:25:24+0200
jkirk@invincible ~ % date +"%Y-%m-%dT%H:%M:%S%Z"
2018-05-24T18:25:40CEST
```

Although I prefer the "+0200" output, I would go for the timezone name option as it is in POSIX.

jkirk: After an internal discussion, we will go for:
```
% date +"%Y%m%d_%H%M%S_%Z"
20180525_101342_CEST
```

moviuro: Why would you dismiss an international standard?

mika: What do you mean with "international standard"?
Our approach was to make sure the filenames work on every filesystem, ':' characters are causing troubles on certain filesystems (as they aren't support on e.g. VFAT).

jkirk: On a VFAT filesystem the following is happening:
```
% bk firmware-iwlwifi_20161130-3_all.deb 
cp: cannot create regular file ‘firmware-iwlwifi_20161130-3_all.deb_2018-05-25T09:07:46Z’: Invalid argument
```

moviuro: See https://en.wikipedia.org/wiki/ISO_8601

For instance, `+"%Y%m%dT%H%M%S%Z"` would be standard compliant + not broken on VFAT.

moviuro: I will have to repair L3263 L3268 L3270 before that.

jkirk: I tryed to fix/test my changes for L3263 L3268 L3270 with something like this:
```
ls README.md_[0-9](#c8)T([0-1][0-9]|2[0-3])([0-5][0-9])(#c2)([A-Z])(#c2,4)                    
README.md_20180601T154447CEST
```
I assumed a timezone abbreviation length between 2 and 4 characters. I looked for the full list and just found this: https://en.wikipedia.org/wiki/List_of_time_zone_abbreviations
which shows time zone abbreviations between 3 and 5 characters. So the question is where can I find the list which is used by date(1)?

I then tested some time zones:
```
TZ=Europe/Vienna date +"%Y%m%dT%H%M%S%Z"
20180601T165353CEST
TZ=America/Santiago date +"%Y%m%dT%H%M%S%Z"
20180601T105402-04
```
To quote @KellerFuchs: 
> What's the issue with using UTC?

Why can't we stick to UTC and just change the format. Any of these two would be fine for me:
```
TZ=America/Santiago date -u +"%Y%m%dT%H%M%S%Z"
20180601T150236UTC
TZ=America/Santiago date -u +"%Y%m%dT%H%M%SZ"
20180601T150637Z
```

What do you think?

jkirk:
@mika Please merge this change, if it ok for you as no feedback were provided till now. Thank you!

KellerFuchs:
@mika As far as I can tell, this should have been closed by a5f5b62, no?

mika:
Jepp, just forgot to close this one, thank everyone!

PR #66, opened 2018-05-13
zshrc: use `apt` instead of `apt-get` also for search + show

mika:
Output of `apt show` and `apt search` is more user-friendly,
so switch to it.

While at it fix description of `au` alias (it's apt and not
apt-get).


mika:
@ft @formorer @jkirk @evgeni - please comment/review :)

ft:
I don't know how this approval thing on github works, but anyway. I
don't see a reason not to do this (I mean sure, the output works a
bit differently, but for the better if you ask me). So just go ahead.

mika:
Thanks folks, merged!

PR #70, opened 2018-07-23
bk(): Remove the separators between the units of time

jkirk:
In PR #60 we discussed the use of the local timezone instead of UTC.
While doing so we figured out that the current implementation is broken
on VFAT.

This solution removes the delimiters between the units of time which is fully
IS0 8601 compliant and should work well on all filesystems.

@moviuro: I did not want to interfere with your PR, so I would be pleased if you can give me any feedback to this change. (Using the local time vs UTC should be discussed in your PR, but removing the separators is needed anyway).

/cc @mika, @btittelbach (please take a look if you have some time).

NOTE: this PR had to be recreated as the one PR #68 was closed automatically because I pushed to grml/grml-etc-core instead into my fork (jkirk/grml-etc-core).

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -3268,7 +3268,7 @@ fi
 #f5# Backup \kbd{file_or_folder {\rm to} file_or_folder\_timestamp}
 function bk () {
     emulate -L zsh
-    local current_date=$(date -u "+%Y-%m-%dT%H:%M:%SZ")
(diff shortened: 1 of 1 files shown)


jkirk:
I thought about that too, but came to the conclusion that this is such a minor thing, that everyone using `bk -r` should be able to run `rm [...]` and find the right backup files to be deleted.

mika:
Rebased + merged, thx!

PR #73, opened 2019-02-11
Disable bracketed paste mode for dumb terminals

sometimesfood:
Starting with zsh version 5.1, ZLE began supporting "bracketed paste mode".

Currently, even on dumb terminals, ZLE sometimes emits the escape sequence "[?2004h" after the prompt if bracketed paste mode is not disabled. This prevents various tools that try to detect a shell prompt (such as Emacs TRAMP) from doing so successfully. (The shell prompt effectively becomes `user@host ~ % [?2004h` – which typically will not be matched by prompt detecting regexps, thereby breaking Emacs TRAMP for target hosts that use the Grml zsh config and zsh ≥5.1.)

Since most "dumb" terminals do not have bracketed paste capabilities anyway, this pull request changes the Grml zsh config to disable bracketed paste mode for "dumb" terminals.

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -3733,6 +3733,9 @@ if check_com -c hg ; then
 
 fi # end of check whether we have the 'hg'-executable
 
+# disable bracketed paste mode for dumb terminals
+[[ "$TERM" == dumb ]] && unset zle_bracketed_paste
+
 # grml-small cleanups and workarounds
 
 # The following is used to remove zsh-config-items that do not work


mika:
Thanks for this PR, @sometimesfood! I just merged and pushed it :)

Feel free to add the text like in this pull request in the body of your commit message next time, that's certainly useful information that would be nice to have in git! (I added it to the commit message of the merge-commit this time :))

/ping @ft to please review this and if there's anything further to be done please let us know, thx! :)

sometimesfood:
@mika: Thanks, will do!

@ft: Do you think this should/could be addressed directly in zsh? zsh already tries to detect Emacs shell buffers and disables ZLE completely if `$TERM == "emacs"` (as documented in the zsh FAQ), but it might make sense to disable bracketed paste mode for all dumb terminals directly in zsh.

ft:
I think this should indeed be addressed in zsh upstream. But it doesn't hurt to have a workaround in place.

PR #77, opened 2019-05-17
tmux: use foo-style to set status styles

zdtcd:
From tmux 1.9, (20 Feb 2014), foo-{bg,fg,attr} commands have been
deprecated and tmux recommended to use foo-style instead.

Those options have been removed from tmux 2.9

It's more than 5 years from the deprecation of foo-{bg,fg,attr}.
it should be safe to use the new option by now.

Signed-off-by: Doan Tran Cong Danh <congdanhqx@gmail.com>

### DIFF
--- a/etc/tmux.conf
+++ b/etc/tmux.conf
@@ -45,8 +45,7 @@ set -g visual-activity on
 set -g mode-keys vi
 
 ### set status line appearance
-set -g status-fg white
-set -g status-bg black
+set -g status-style fg=white,bg=black
 
 set -g status-left-length 28
 ### status-left:  @hostname:sessionname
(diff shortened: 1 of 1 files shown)


mika:
Thanks for this PR, @congdanhqx!

The change LGTM, given that tmux 1.9-6 is in oldstable I *think* we should be safe with accepting this. Any feedback from other tmux users? @grml/team-grml-developers

evgeni:
Untested, but the explanation is reasonable and I don't care about anything older than stable ;)

congdanhqx:
Forget to mention that this setting was tested with tmux 2.9a on Void Linux, and Arch Linux.

2.9 is still in Debian experimental, I think it may go to sid when buster become the new stable.

jkirk:
Just tested this change with tmux v2.3-4 (Debian/stretch) and all seems to work as expected.

@mika: +1 from my side.

mika:
Thanks everyone, merged!

PR #78, opened 2019-06-13
Update style options to be tmux 2.9 comptible

joech4n:
Per https://github.com/tmux/tmux/wiki/FAQ#how-do-i-translate--fg--bg-and--attr-options-into--style-options
> In tmux 1.9 each set of three options were combined into a single option (so mode-fg, mode-bg and mode-attr became mode-style) and in tmux 2.9 the old options were removed.

From https://github.com/tmux/tmux/issues/1689#issuecomment-486722349
> The form is exactly the same, it is just one option instead of three:
> 
> `set -g mode-style bg=red,fg=green,blink`

Tested on tmux 2.9a (OSX)

### DIFF
--- a/etc/tmux.conf
+++ b/etc/tmux.conf
@@ -45,8 +45,7 @@ set -g visual-activity on
 set -g mode-keys vi
 
 ### set status line appearance
-set -g status-fg white
-set -g status-bg black
+set -g status-style bg=black,fg=white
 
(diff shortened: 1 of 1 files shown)


jkirk:
@joech4n: Thx for your contribution. This is actually a duplicate of PR #77.

Issue #74, opened 2019-02-27
Multiline command re-echoing when using urxvt


cddmp:
When running multiline commands in urxvt, parts of the command are re-echoed. 

Example:
````
me@system ~/foo % ls \
\`> -l

-ltotal 0
-rw-r--r-- 1 me users 0 Feb 27 20:36 bar1
-rw-r--r-- 1 me users 0 Feb 27 20:36 bar2
me@system ~/foo %
````
Note the '-l' be re-echoed in front of 'total'.

After some debugging it turned out that the problems is in the function info_print() which is called by set_title(). urxvt seems to wrongly handle the newline character. A fix would be to replace line 2484 in zshrc which is 

`printf '%s' "$*"`

with

`printf '%s' "${*//[$'\n']}"`

Since this problem seems to be limited to urxvt, this might be something which need to be actually fixed in urxvt. I just wanted to bring it up here. Others seem to have come across the same issue:
https://stackoverflow.com/questions/16120888/grmls-zsh-config-multiline-command-re-echoing

mika:
Thanks for reporting this, @cddmp! @ft are you aware of anything that we could do about it?

ft:
Will try to look at it this weekend. Can't promise, though.

cantandwont:
I used zsrc.local to redefine the function `grml_control_xterm_title` from [zshrc](https://github.com/grml/grml-etc-core/blob/ffe57c8786bf71c015d948eff4253089725b9f8e/etc/zsh/zshrc#L2525), changing
`set_title "${(%):-"%n@%m:"}" "$1"`
to
`set_title "${(%):-"%n@%m:"}" "${(z)1}"`

cantandwont:
Tried out Alacritty - it trims from the window title anything after (and incl.) the first newline or semicolon, whether it's quoted or not (eg. when running `echo "12;3" | ...` window title has `echo "12`).
Alacritty had no problem with re-echoing lines after the prompt.
For now, I'm going with urxvt and `set title` ... `"${1%%[[:cntrl:]]*}"` in zshrc

cddmp:
@cantandwont: Your solution trims everything after the newline character, the solution I presented above does not.

cantandwont:
[...]

PR #80, opened 2019-07-15
Update zshrc - Fix urxvt multi-line issue

cantandwont:
resolves issue #74


mika:
We squashed your commits, adjusted the commit message and just pushed your change.
Thanks for the PR and fix!

PR #81, opened 2019-07-16
Fix multiline command re-echoing in urxvt

cddmp:
My attempt to solve #74 - which works without trimming the title.

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -2481,7 +2481,7 @@ function info_print () {
     esc_end="$2"
     shift 2
     printf '%s' ${esc_begin}
-    printf '%s' "$*"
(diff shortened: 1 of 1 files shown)


mika:
Thanks for the PR, but this seems to be more intrusive than the fix we just pushed in commit 0c9e97b682. Please let us know if there's anything unsolved.

cddmp:
As I wrote in #74, my PR was meant as a blocker to not silently merge cantandwont's first PR which I think was not a good solution. I didn't issue a PR in the first place, since I thought you guys come up with a better fix than my dirty workaround. I'm more than happy cantandwont digged deeper into this and finally found an easy and great solution. Thank you :)

mika:
@cddmp very much appreciated, thanks also from our side for your efforts!

PR #82, opened 2019-07-20
zshrc: Reset $REPLY parameter before calling prompt token functions

dmitrivereshchagin:
It may be tempting to write a prompt token function like this

    if (( ${+commands[kerl]} )) REPLY=$(kerl prompt '<%s> ')

But if the command does not exist user may be end up with duplicated
tokens in her prompt.  Resetting `$REPLY` parameter before calling
user-defined functions prevents this problem.


mika:
Thanks for the fix, @dmitrivereshchagin!

@ft please let us know if this change might cause any issues we aren't aware of, thanks!

PR #84, opened 2019-08-16
screenrc: Add 'xterm-256color' to terminals

LeSpocky:
KDE's Konsole terminal sets env variable TERM to 'xterm-256color' by
default in newer versions, so the scrollbar fix did not match anymore.

---

This actually annoyed me for a long time now, and today I decided to do my research. I found the original solution at http://www4.cs.fau.de/~jnweiger/screen-faq.html and then realized it was all right there in grml's screenrc, only missing the little piece from this commit.


mika:
Nice catch, thanks for the PR, just merged it!

PR #87, opened 2019-11-14
zshrc: support PAGER='less -Mr' usage and fall back to vi if LESSOPEN is unset

mika:
It's horribly annoying when PAGER is set to less, but lesspipe isn't configured
via LESSOPEN, then `dchange $package` fails with less like this:

% dchange zsh
"/usr/share/doc/zsh/changelog.Debian.gz" may be a binary file.  See it anyway?

vi(m) handles *.gz just fine, so let's try to use vi if PAGER is set to
less and LESSOPEN is unset.

While at it fix usage of ${PAGER} vs ${=PAGER} to also handle
a setting like PAGER='less -Mr' (otherwise failing with
`command not found: less -Mr`).

That's the best I could come up with for now without using vi by default and still supporting $PAGER. Any comments/better ideas?

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -2928,9 +2928,6 @@ function sll () {
     return ${RTN}
 }
 
-# TODO: Is it supported to use pager settings like this?
-#   PAGER='less -Mr' - If so, the use of $PAGER here needs fixing
-# with respect to wordsplitting. (ie. ${=PAGER})
 if check_com -c $PAGER ; then
     #f3# View Debian's changelog of given package(s)
     function dchange () {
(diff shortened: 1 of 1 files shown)


etc/zsh/zshrc
```diff
@@ -2928,23 +2928,28 @@ function sll () {
...
     #f3# View Debian's changelog of given package(s)
     function dchange () {
         emulate -L zsh
         [[ -z "$1" ]] && printf 'Usage: %s <package_name(s)>\n' "$0" && return 1
 
         local package
+
+        # less without e.g. `|lesspipe %s` inside $LESSOPEN can't properly read *.gz files,
+        # try to use vi(m) instead
+        local viewer="${=PAGER}"
```
akorn:
```
local viewer
if [[ ${$(typeset -p PAGER)[2]} = -a ]]; then viewer=($PAGER); else viewer=(${=PAGER}); fi
```
This has the advantage of properly handling a PAGER variable declared as an array without splitting its elements into words.

In general, it is better to assemble shell commands in array variables than in scalars because handling of embedded spaces is more intuitive and straightforward.

etc/zsh/zshrc
```diff
@@ -2928,23 +2928,28 @@ function sll () {
...
+        if [[ "${PAGER}" == "less" ]] && [[ -z "${LESSOPEN}" ]] && check_com vi ; then
+          viewer='vi'
+        fi
+
         for package in "$@" ; do
             if [[ -r /usr/share/doc/${package}/changelog.Debian.gz ]] ; then
-                $PAGER /usr/share/doc/${package}/changelog.Debian.gz
+                ${=viewer} /usr/share/doc/${package}/changelog.Debian.gz
```
akorn: Instead of `${=viewer}` I suggest `$viewer`, with `viewer` being an array.

ft:
LGTM

mika:
Thanks everyone, appreciate your feedback! :+1:

PR #88, opened 2019-11-14
zsh: commit-to-history: keep escaped character

sgn:
Prior to this change, when committing this line to history

    echo "hello\"world"

zsh will record this line in history instead:

    echo "hello"world"

because the \" is processed by print.

Correct it by print in raw mode.

Signed-off-by: Doan Tran Cong Danh <congdanhqx@gmail.com>


mika:
Nice catch, thanks for providing a PR! :+1:

PR #95, opened 2020-05-22
zshrc:  Fix darwin battery info field selection

gholms:
The batterydarwin function looks for the info it wants using field
numbers that are off by one on the current version of Mac OS.  I'm not
sure which version this changed in or what exactly they added (I suspect
it is the battery ID), but this patch makes battery info start working
again.

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -1856,7 +1856,7 @@ done
 function batterydarwin () {
 GRML_BATTERY_LEVEL=''
 local -a table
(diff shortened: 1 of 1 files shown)


mika:
Thanks for the PR, @gholms! Which Mac OS version are you using?
@zeha any chance that you might check whether this works for you as well?

zeha:
WFM on Catalina

mika:
Thanks for testing, @zeha and thanks for the PR, @gholms!

Issue #96, opened 2020-05-24
zshrc: warning, when hostname is not found


apt-ghetto:
I am using Arch Linux with systemd and have downloaded the .zshrc config from here.

It was shown a warning, that "hostname" was not found, which is true, because it is not installed. And because i did not want to install another package, I have "fixed" this on my machine by replacing
```
$(hostname)
```
with
```
$(hostnamectl --static)
```

I would appreciate, if you could add a check for `hostname` and `hostnamectl`.

mika:
@apt-ghetto does https://github.com/grml/grml-etc-core/pull/97 fix the issue for you?
@ft @jkirk  I'd appreciate if you could briefly check whether https://github.com/grml/grml-etc-core/pull/97 makes sense for you,  thanks :)

apt-ghetto:
@mika Yes, it is working for me. Thank you.

jkirk:
This might be a silly question, but is it possible to have non of both?

mika:
> This might be a silly question, but is it possible to have non of both?

Could be, though have never seen that myself and before introducing a workaround (like falling back to `/etc/hostname`) I would like to know about it :)

moviuro:
`uname -n` looks like the way to go. It's [actually POSIX](https://pubs.opengroup.org/onlinepubs/009695399/utilities/uname.html)

PR #97, opened 2020-05-26
zshsrc: support hostnamectl iff hostname isn't present

mika:
The hostname binary doesn't necessarily exist, while
hostnamectl is available.

Closes: grml/grml-etc-core#96


ft:
LGTM

mika:
Thanks for reporting, @apt-ghetto and thanks for the review, @ft - highly appreciated!

PR #100, opened 2020-05-28
zshenv, zshrc: add fallback to uname for HOSTNAME

moviuro:
`uname -n` is POSIX, and should work when neither hostname(1) nor hostnamectl(1) exist.

Enhances #97, see #96

### DIFF
--- a/etc/zsh/zshenv
+++ b/etc/zsh/zshenv
@@ -32,6 +32,8 @@ elif [[ -x $(which hostname) ]] ; then
   export HOSTNAME="$(hostname)"
 elif [[ -x $(which hostnamectl) ]] ; then
   export HOSTNAME="$(hostnamectl --static)"
+else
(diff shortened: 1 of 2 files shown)


mika:
Oh nice, thanks! :+1:

jkirk:
And here we go with the next (possibly) silly question: Why not use `uname -n` as primary and possibly only source for HOSTNAME?

mika:
> And here we go with the next (possibly) silly question: Why not use `uname -n` as primary and possibly only source for HOSTNAME?

Hm, any counter-arguments against using `uname -n` by default? @ft @moviuro @zeha

zeha:
s/by default//;

i think this should really just become:
```
if [ -n "${HOSTNAME}" ] ; then
  export HOSTNAME="${HOSTNAME}"
else
   export HOSTNAME="$(uname -n)"
 fi
```

Issue #99, opened 2020-05-27
vimrc: skip_defaults_vim not available with vim-tiny


mika:
On a custom Grml ISO build as of today with most recent grml-etc-core and vim-tiny package only (so not vim/vim-runtime)
```
% vim /some/file
Error detected while processing /usr/share/vim/vimrc:
line   30:
E319: Sorry, the command is not available in this version: let g:skip_defaults_vim=1
% ls -la /usr/share/vim/vimrc
lrwxrwxrwx 1 root root 14 May 12 02:37 /usr/share/vim/vimrc -> /etc/vim/vimrc
```

mika:
@jkirk I'd appreciate review of https://github.com/grml/grml-etc-core/commit/3703c8070fc472cc4e906caf3b5aec70c0c3aed1

mika:
@jkirk https://github.com/grml/grml-etc-core/pull/101 is addressing this

PR #101, opened 2020-05-28
vimrc: support vim-tiny and older versions of vim

mika:
If vim-tiny is present (lacking +eval feature, see `vim --version`),
then invoking vim fails with:

| Error detected while processing /usr/share/vim/vimrc:
| line   30:
| E319: Sorry, the command is not available in this version: let g:skip_defaults_vim=1

The file /usr/share/vim/vimrc is a symlink to /etc/vim/vimrc:

| % ls -la /usr/share/vim/vimrc
| lrwxrwxrwx 1 root root 14 May 12 02:37 /usr/share/vim/vimrc -> /etc/vim/vimrc

So if the eval feature is missing, don't set skip_defaults_vim=1.

Furthermore, on older vim versions we're failing with:

| Error detected while processing /code/etc/vim/vimrc:
| line   61:
| E474: Invalid argument: display=truncate
| line  201:
| E474: Invalid argument: listchars=eol:$,precedes:«,extends:»,tab:»·,trail:·

The "set display=truncate" only works as of patch-7.4.2115 or newer,
see upstream vim.git commit b9a46fec3e79.

The listchars setting only works when running in utf-8 mode,
by default vim-tiny is running in latin1 though.

Closes: grml/grml-etc-core#99

### DIFF
--- a/etc/vim/vimrc
+++ b/etc/vim/vimrc
@@ -27,7 +27,11 @@ runtime! debian.vim
 " | Do *not* source $VIMRUNTIME/defaults.vim (in case no user vimrc is found),
 " | since that prevents overwriting values we want to set in this file,
 " | instead we set settings similar to what can be found in default.vim below
-let g:skip_defaults_vim=1
+" |
+" | Only do this part when Vim was compiled with the +eval feature.
+if 1
+  let g:skip_defaults_vim=1
+endif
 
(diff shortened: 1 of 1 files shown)


jkirk:
Just tested vim-tiny in a minimal Debian/jessie and Debian/bullseye docker environment. LGTM. :+1:

PR #102, opened 2020-05-30
modified: etc/zsh/zshrc, add zstd support for simple-extract().

ur4t:
More distros move to zst as pkg compression format (such as archlinux), so add zst support


mika:
Oh excellent idea, thanks for your PR, merged! :+1:

Issue #85, opened 2019-08-21
alias url-quote is broken


jkirk:
https://github.com/grml/grml-etc-core/blob/24e8d3be9f59a015d4292b09beb38ea707a260be/etc/zsh/zshrc#L1577

```sh
jkirk@executor ~ % echo https://github.com/grml/grml-etc-core/blob/24e8d3be9f59a015d4292b09beb38ea707a260be/etc/zsh/zshrc#L1577
zsh: no matches found: https://github.com/grml/grml-etc-core/blob/24e8d3be9f59a015d4292b09beb38ea707a260be/etc/zsh/zshrc#L1577
jkirk@executor ~ % url-quote                                                      
jkirk@executor ~ % echo https://github.com/grml/grml-etc-core/blob/24e8d3be9f59a015d4292b09beb38ea707a260be/etc/zsh/zshrc#L1577
zsh: no matches found: https://github.com/grml/grml-etc-core/blob/24e8d3be9f59a015d4292b09beb38ea707a260be/etc/zsh/zshrc#L1577
```

(Please note: you have to paste the URL after each `echo`. Do not use your history to reuse the command.)

According to https://github.com/sorin-ionescu/prezto/issues/978 and https://github.com/zsh-users/zsh/blob/master/Functions/Zle/url-quote-magic#L11 the fix is simple:

```sh
alias url-quote='autoload -Uz url-quote-magic bracketed-paste-magic; zle -N self-insert url-quote-magic; zle -N bracketed-paste bracketed-paste-magic'
```

With this the URL is quoted automatically:
```
% url-quote
% echo https://github.com/grml/grml-etc-core/blob/24e8d3be9f59a015d4292b09beb38ea707a260be/etc/zsh/zshrc\#L1577
https://github.com/grml/grml-etc-core/blob/24e8d3be9f59a015d4292b09beb38ea707a260be/etc/zsh/zshrc#L1577
```

mika:
@ft are you aware of any reason why url-quote shouldn't be loaded/active by default (so we don't even need the url-quote alias at all)?

@jkirk thanks for spotting this! \o/

PR #104, opened 2020-06-03
zshrc: url-quote-magic depends on bracketed-paste-magic since zsh 5.1

mika:
url-quote with zsh >=5.1 does no longer work as such, it depends
on bracketed-paste-magic to work as it used to do

See https://github.com/zsh-users/zsh/blob/68405f31a043bdd5bf338eb06688ed3e1f740937/README#L38-L45
and https://archive.zhimingwang.org/blog/2015-09-21-zsh-51-and-bracketed-paste.html#update
and https://github.com/sorin-ionescu/prezto/issues/978

/cc @ft @btittelbach @jkirk I'd appreciate a review whether this makes sense as such, thanks!

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -113,6 +113,11 @@ fi
 
 # check for version/system
 # check for versions (compatibility reasons)
+function is51 () {
+    [[ $ZSH_VERSION == 5.<1->* ]] && return 0
+    return 1
+}
+
 function is4 () {
     [[ $ZSH_VERSION == <4->* ]] && return 0
     return 1
@@ -1584,7 +1589,13 @@ zrcautoload zed
 # else
 #    print 'Notice: no url-quote-magic available :('
 # fi
-alias url-quote='autoload -U url-quote-magic ; zle -N self-insert url-quote-magic'
+if is51 ; then
+  # url-quote doesn't work without bracketed-paste-magic since Zsh 5.1
+  alias url-quote='autoload -U bracketed-paste-magic url-quote-magic;
+                   zle -N bracketed-paste bracketed-paste-magic; zle -N self-insert url-quote-magic'
+else
+  alias url-quote='autoload -U url-quote-magic ; zle -N self-insert url-quote-magic'
+fi
 
 #m# k ESC-h Call \kbd{run-help} for the 1st word on the command line
 alias run-help >&/dev/null && unalias run-help


jkirk:
Cool, this finally closes #85. I just test this change with zsh in Debian/buster and it works fine. Will try it on the older machines too...

jkirk:
Just tested this change in my minimal Debian/jessie, Debian/stretch and Debian/bullseye docker environment. LGTM! :+1: 

One note though: When using the default `/etc/zsh/zshrc` configuration file, pasting the URLs works fine *without* `url-quote` aka `url-quote-magic`. How can one test if a zsh function is loaded?

And: How can I always enable `url-magic` and how to disable / unload it? (We should document that somewhere.)

mika:
> Just tested this change in my minimal Debian/jessie, Debian/stretch and Debian/bullseye docker environment. LGTM! +1

Excellent :)

> One note though: When using the default `/etc/zsh/zshrc` configuration file, pasting the URLs works fine _without_ `url-quote` aka `url-quote-magic`. How can one test if a zsh function is loaded?

Not for me?

```
% wget https://github.com/search?q=magic-abbrev+zshrc&type=Commits
[1] 2692
zsh: no matches found: https://github.com/search?q=magic-abbrev+zshrc                                                                                                                                                                        
[1]  + 2692 exit 1     wget https://github.com/search?q=magic-abbrev+zshrc
```

Whereas with:

```
% url-quote
% wget https://github.com/search\?q\=magic-abbrev+zshrc\&type\=Commits
--2020-06-03 13:30:59--  https://github.com/search?q=magic-abbrev+zshrc&type=Commits
Resolving github.com (github.com)... 140.82.118.3
[...]
2020-06-03 13:31:00 (1.67 MB/s) - 'search?q=magic-abbrev+zshrc&type=Commits' saved [226819]
```

> And: How can I always enable `url-magic` and how to disable / unload it? (We should document that somewhere.)

Indeed, would be nice to have a simple toggling feature for usage from within e.g. `~/.zshrc.local`, yes.
I'm afraid the fact that it needs to be invoked manually via `url-quote` actually lead to no one using as such. :-/

jkirk:
> Not for me?
> 
> ```
> % wget https://github.com/search?q=magic-abbrev+zshrc&type=Commits
> [1] 2692
> zsh: no matches found: https://github.com/search?q=magic-abbrev+zshrc                                                                                                                                                                        
> [1]  + 2692 exit 1     wget https://github.com/search?q=magic-abbrev+zshrc
> ```

Ah, there seems to be more magic to check for:
```
% docker pull debian:bullseye
bullseye: Pulling from library/debian
Digest: sha256:0b674f8c265556ce7ddfb8c5ce304eec4c771046251d6da52f58135d3fbf0c23
Status: Image is up to date for debian:bullseye
docker.io/library/debian:bullseye

% docker run -i -t debian:bullseye
root@54004e9b6830:/# apt update
[...]

root@54004e9b6830:/# apt install zsh

54004e9b6830# wget https://github.com/search?q=magic-abbrev+zshrc&type=Commits
[1] 211
54004e9b6830# zsh: no matches found: https://github.com/search?q=magic-abbrev+zshrc

[1]  + exit 1     wget https://github.com/search?q=magic-abbrev+zshrc
```

This is the URL I have tested which fails in my system but works OOB:
```
54004e9b6830# echo https://github.com/grml/grml-etc-core/blob/24e8d3be9f59a015d4292b09beb38ea707a260be/etc/zsh/zshrc#L1577
https://github.com/grml/grml-etc-core/blob/24e8d3be9f59a015d4292b09beb38ea707a260be/etc/zsh/zshrc#L1577
```
Let us assume `url-quote` does the trick.

(Defined) functions can be checked like that:
```
% functions | grep "^url-quote"
url-quote-magic () {
```

I tried `unfunction url-quote-magic` but then zsh complains with **No such shell function `url-quote-magic'** when I try to paste something.

jkirk:
JFTR, this seems to revert `url-quote`:
```
% zle -D self-insert bracketed-paste
```

jkirk:
Ping @ft @btittelbach. What do you think? Are you ok with this PR? Thx!

ft:
Not an url-quote magic user, so can't comment on the user interface. The code looks fine.

Issue #107, opened 2020-06-24
RFC for zshrc: disable smiley on failures in right side of shell prompt by default


mika:
I'm proposing to disable our smiley which appears if exit code != 0 on the right side of the prompt, like:

```
mika@lunge /tmp/tmp.kvMVtGqka4 % false
1 mika@lunge /tmp/tmp.kvMVtGqka4 %                                                          :(
```

IMO it's totally annoying for c/p.
I'm aware that users can control this behavior via:

```
# disable smiley
zstyle ':prompt:grml:right:setup' use-rprompt false
```

... in their personal zsh configuration, but I'd prefer to disable this by default, so folks don't have to manually remove it in c/p for docs etc.

If someone objects, please speak up! :)

mika:
Ping @formorer @ft @jkirk @btittelbach @zeha for comments/opinions :)

zeha:
Please do it, I've always turned this off in my config copies.

PR #108, opened 2020-06-25
zshrc: disable sad-smiley in rprompt by default, making c/p harder than necessary

mika:
To re-enable it for the default grml prompt use:

  zstyle ':prompt:grml:right:setup' items sad-smiley

Or when using the grml-large prompt theme, to enable it:

  zstyle ':prompt:grml-large:right:setup' items sad-smiley

Closes: grml/grml-etc-core#107

I'm not sure whether I found the best way how to achieve this (to not use sad-smiley by default any longer, but making it easy to re-enable it for users who like that), I'd very much appreciate thorough review from @ft :)

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -2016,14 +2016,19 @@ function prompt_grml_help () {
 
     The styles:
 
-        - use-rprompt (boolean): If \`true' (the default), print a sad smiley
-          in $RPROMPT if the last command a returned non-successful error code.
+        - use-rprompt (boolean): If \`true' (the default), then print the defined
+          items (empty by default) in $RPROMPT.
+          For example to enable the sad-smiley in $RPROMPT using the grml prompt
+          this can be achieved via:
+
+             zstyle ':prompt:grml:right:setup' items sad-smiley
+
           (This in only valid if <left-or-right> is "right"; ignored otherwise)
 
         - items (list): The list of items used in the prompt. If \`vcs' is
           present in the list, the theme's code invokes \`vcs_info'
           accordingly. Default (left): rc change-root user at host path vcs
-          percent; Default (right): sad-smiley
+          percent; Default (right): none (being unset)
 
         - strip-sensitive-characters (boolean): If the \`prompt_subst' option
           is active in zsh, the shell performs lots of expansions on prompt
@@ -2078,7 +2083,7 @@ function prompt_grml-large_help () {
 
         - left: rc jobs history shell-level change-root time date newline user
                 at host path vcs percent
-        - right: sad-smiley
+        - right: none (being unset)
 __EOF0__
 }
 
@@ -2383,7 +2388,7 @@ function prompt_grml_precmd () {
     local grmltheme=grml
     local -a left_items right_items
     left_items=(rc change-root user at host path vcs percent)
-    right_items=(sad-smiley)
+    right_items=()
 
     prompt_grml_precmd_worker
 }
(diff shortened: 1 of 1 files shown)


ft:
The way you're doing it is fine and I don't lean either way (enabled per default or not).

One question: Is enabling the `transient_rprompt` option good enough to fix your copy'n'paste problems? With the rprompt behaviour remaining at our previous default?

ziQ:
Awesome. This annoyed me for years, but I was never annoyed enough to figure out how to fix it ;-)

Thanks a lot!

mika:
@ft  thanks for your prompt aeh fast response! :)

> One question: Is enabling the transient_rprompt option good enough to fix your copy'n'paste problems?
> With the rprompt behaviour remaining at our previous default?

Do you mean the `zstyle ':prompt:grml:right:setup' use-rprompt false`?
I personally don't mind too much how we get there, though had to research myself what needs to be done and this was my best solution so far. 8-)

What I try to achieve here:

* avoid breaking the rprompt behavior of existing setups (if folks had something else than `sad-smiley` configured with it)
* make it easy to get the previous behavior back (re-enable the smiley if someone finds this useful)

What would be the best way to get there from your PoV?

ft:
Michael Prokop wrote:
>> One question: Is enabling the transient_rprompt option good enough to fix your copy'n'paste problems?
>> With the rprompt behaviour remaining at our previous default?
>
> Do you mean the `zstyle ':prompt:grml:right:setup' use-rprompt false`?
>
> I personally don't mind too much how we get there, though had to
> research myself what needs to be done and this was my best solution so
> far. 8-)

No, I mean the `transient_rprompt` option. You'd set it like this at the
command prompt:

    setopt transient_rprompt

From the manual:

    TRANSIENT_RPROMPT
      Remove  any  right  prompt from display when accepting a command
      line.  This may be useful with terminals  with  other  cut/paste
      methods.

> What I try to achieve here:
>
> * avoid breaking the rprompt behavior of existing setups (if folks had
>   something else than `sad-smiley` configured with it)
> * make it easy to get the previous behavior back (re-enable the smiley
>   if someone finds this useful)
>
> What would be the best way to get there from your PoV?

I think  what is problematic  is when there  is text in  RPOMPT position
with past-prompts. It doesn't matter for the *currently* *active* prompt
since there's no output after it yet.

That is what transient_rprompt cures: It  removes any such text, in past
prompts being displayed, enabling easy copy'n'pasting.

I think setting that  option would be good enough for  _me_. But I don't
use rprompts  myself, soooo, I  wonder if it's  good enough for  you. If
it's not, your change does what you want, and in the right waym too.

ft:
Here is what I said about this a second ago on IRC:

    ft » "setopt transient_rprompt" should enable easy copy-and-paste with right hand side prompts active.
    ft » Looks like we had that set in the past.
    ft » But I seem to have removed it in 08ab5d927b758f7f3445d0dd57b6f9076d1636ca when I rewrote the prompt code 7 years ago.
    ft » I never heard any complaints about this before.
    ft » -is41 && setopt transient_rprompt
    ft » ^- that's the offending line from the diff.

If you take a look at 08ab5d927b758f7f3445d0dd57b6f9076d1636ca you'll see the removal.

So, fixing this is just adding that line again:

    is41 && setopt transient_rprompt

Like I said, I didn't hear this complaint in 7 years. Would have been an easy fix all along.

jkirk:
@ft Nice feature! With this `transient_rprompt` the sad-smiley is shown once but does not clutter the terminal. I took a screencast to demonstrate this, maybe this is the compromise we are looking for?:

grml-zsh-smiley

mika:
Oh, what a great find, @ft \o/ Many thanks!

I prepared another PR for this: https://github.com/grml/grml-etc-core/pull/109

mika:
Was replaced via https://github.com/grml/grml-etc-core/pull/109, closing hereby

PR #110, opened 2020-07-04
zshrc: set title in alacritty

sbruder:
[Alacritty](https://github.com/alacritty/alacritty) is a “cross-platform, GPU-accelerated terminal emulator” written in Rust. By default, its window title is “Alacritty”. Since its `TERM` value is `alacritty`, zsh with the grml zshrc does not set the window title. This PR adds `alacritty` to the supported `TERM`s for setting the window title.


mika:
Thanks for the PR! :+1:

PR #115, opened 2020-11-14
README.md: Add tmux.conf to the "most important files" list

jkirk:
Please add tmux.conf to the "most important files" list.

And while at it, please update [grml-config.sh](https://grml.org/scripts/grml-config.sh) too. Thx!


mika:
Good idea, thanks!

PR #116, opened 2020-11-25
Change "command ls" to "$(builtin whence -p ls)" in zshrc

zeertzjq:
When I have an alias `sudo="sudo "` and `command ls` is used in the `ls` aliases, these aliases cannot be used with `sudo` because `command` is a zsh builtin command. Changing `command ls` to `$(builtin whence -p ls)` for zsh and `$(which ls)` for bash can solve this problem.

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -2663,23 +2663,24 @@ if check_com -c screen ; then
     fi
 fi
 
+LS="$(builtin whence -p ls)"
 # do we have GNU ls with color-support?
 if [[ "$TERM" != dumb ]]; then
     #a1# List files with colors (\kbd{ls \ldots})
-    alias ls="command ls ${ls_options:+${ls_options[*]}}"
+    alias ls="$LS ${ls_options:+${ls_options[*]}}"
     #a1# List all files, with colors (\kbd{ls -la \ldots})
-    alias la="command ls -la ${ls_options:+${ls_options[*]}}"
+    alias la="$LS -la ${ls_options:+${ls_options[*]}}"
     #a1# List files with long colored list, without dotfiles (\kbd{ls -l \ldots})
-    alias ll="command ls -l ${ls_options:+${ls_options[*]}}"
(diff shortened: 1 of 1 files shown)


mika:
@zeertzjq can you please squash your commits into one single commit?
@ft what do you think of this change, with your grml-zshrc hat on? :)

zeertzjq:
> @zeertzjq can you please squash your commits into one single commit?

Done.

mika:
@zeertzjq thanks for squashing. IIRC I was annoyed by some behavior regarding sudo as well, though sadly can't remember the details. I'm not sure I understand your use case either :) What's the reason for doing `alias sudo="sudo "` in first place? Do you have any steps for us to reproduce this issue so we can easier decide about it?

zeertzjq:
@mika According to the [zsh man pages](https://linux.die.net/man/1/zshbuiltins), when defining aliases:

> alias [ {+|-}gmrsL ] [ name[=value] ... ]
> For each name with a corresponding value, define an alias with that value. A trailing space in value causes the next word to be checked for alias expansion.

And bash seems to have the same feature.

So with alias `sudo="sudo "` present, bash or zsh can expand the alias `la` in when `sudo la` is typed, currently this expands to `sudo command ls -la --color=auto`, which causes an error `sudo: command: command not found` because `command` is not an executable but a shell builtin. This pull request makes `sudo la` expand to `sudo /bin/ls -la --color=auto` or `sudo /usr/bin/ls -la --color=auto` etc., which can be run successfully.

mika:
@ft any objections from your side against merging this? :)

jkirk:
No idea if this affects this PR somehow, but I just read (again) that `which` is not POSIX and should be avoided (see: https://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then).

zeertzjq:
I've also found that `which` may not be present in some Linux systems (e.g. a fresh minimal Arch Linux install), so I'll limit this PR to `zshrc`.

zeertzjq:
Closed due to inactivity.

PR #119, opened 2021-05-17
zshrc: change sudo-command-line switch 'sudo'

a-wing:
fix add 'sudo' plugin zsh-syntax-highlighting error

If not `sudo`, use `sudo-command-line` add

```sh
ls

# sudo-command-line

sudo ls
```

If has `sudo`, use `sudo-command-line` remove

```
sudo ls

# sudo-command-line

ls 
```

If use zsh plugin `zsh-syntax-highlighting`, run `sudo-command-line` this color error. need `reset-prompt`

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -1201,9 +1201,9 @@ zle -N grml-zsh-fg
 function sudo-command-line () {
     [[ -z $BUFFER ]] && zle up-history
     local cmd="sudo "
-    if [[ ${BUFFER:0:${#cmd}} == ${cmd} ]]; then
+    if [[ ${BUFFER} == ${cmd}* ]]; then
         CURSOR=$(( CURSOR-${#cmd} ))
-        BUFFER="${BUFFER:${#cmd}}"
+        BUFFER="${BUFFER#$cmd}"
     else
         BUFFER="${cmd}${BUFFER}"
         CURSOR=$(( CURSOR+${#cmd} ))
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -1200,10 +1200,15 @@ zle -N grml-zsh-fg
 # run command line as user root via sudo:
 function sudo-command-line () {
     [[ -z $BUFFER ]] && zle up-history
-    if [[ $BUFFER != sudo\ * ]]; then
-        BUFFER="sudo $BUFFER"
-        CURSOR=$(( CURSOR+5 ))
+    local cmd="sudo "
+    if [[ ${BUFFER:0:${#cmd}} == ${cmd} ]]; then
+        CURSOR=$(( CURSOR-${#cmd} ))
+        BUFFER="${BUFFER:${#cmd}}"
+    else
+        BUFFER="${cmd}${BUFFER}"
+        CURSOR=$(( CURSOR+${#cmd} ))
     fi
+    zle reset-prompt
 }
 zle -N sudo-command-line
 


mika:
Thanks for the PR! No objections per sē , but sadly I don't understand / can't parse the commit message. Could you try rewording the explanation in the commit message for the underlying problem? Do you have steps to reproduce the issue for us?

a-wing:
I Change two code:
1. I added to remove prefix `sudo`
2. fix syntax-highlighting error


Default:

```sh
bindkey | grep sudo
"^Os" sudo-command-line
```

use this hotkey, auto command add/remove prefix `sudo`

* * *

syntax-highlighting error

Screen Shot 2021-05-17 at 23 30 08

The `reset-prompt` fix this problem. But this is related to the use of themes and plugins

etc/zsh/zshrc
```diff
@@ -1200,10 +1200,15 @@ zle -N grml-zsh-fg
 # run command line as user root via sudo:
 function sudo-command-line () {
     [[ -z $BUFFER ]] && zle up-history
-    if [[ $BUFFER != sudo\ * ]]; then
-        BUFFER="sudo $BUFFER"
-        CURSOR=$(( CURSOR+5 ))
+    local cmd="sudo "
+    if [[ ${BUFFER:0:${#cmd}} == ${cmd} ]]; then
```
ft, line 1204: This looks like a less readable version of  `if [[ $BUFFER == ${cmd}* ]]; then`. I'd prefer the more readable approach.

The `${param:...:...}` indexing syntax is a fairly new addition in zsh. So using the other approach makes the code more portable across zsh versions.

etc/zsh/zshrc
```diff
@@ -1200,10 +1200,15 @@ zle -N grml-zsh-fg
...
+        CURSOR=$(( CURSOR-${#cmd} ))
+        BUFFER="${BUFFER:${#cmd}}"
```
ft, line 1206: So, this is a new feature, with this, you can toggle whether or not `$BUFFER` has a `sudo` prefix. That's the intention?

I like that.

But again, the indexing mechanism used can be changed to something that's more portable across zsh versions: `BUFFER="${BUFFER#$cmd}"`. So I'd prefer that.

etc/zsh/zshrc
```diff
@@ -1200,10 +1200,15 @@ zle -N grml-zsh-fg
...
+    else
+        BUFFER="${cmd}${BUFFER}"
+        CURSOR=$(( CURSOR+${#cmd} ))
     fi
+    zle reset-prompt
```
ft, line 1211: This is for fixing the zsh-syntax-highlight issue, right?

Normally, I would say that there's a problem in zsh-syntax-highlight, because it's its job to correctly put up colour if a widget changes the context if the line-editor's buffer. On the other hand, the impact in negligible and it would work around a problem in exiting zsh-syntax-highlight installations.

Soooo, I suppose we can allow this.

ft (changes requested):

ft:
I do like the use of `${#cmd}` instead of a magic number like `5`. :)

etc/zsh/zshrc
a-wing: This might be better

a-wing:
Remove use indexing syntax

ft:
This new version of the code look good to me! Thanks!

ft (approved):

mika:
Thanks a lot @a-wing for this PR! Thanks @ft for your feedback and review!
:tada:

PR #124, opened 2021-08-23
remove dependency of hostname[ctl]

bhoppi:
Futher improve #105 . There is another place using hostname[ctl].


mika:
Ah nice catch, thank you!

PR #127, opened 2021-10-27
zshrc: use ip from iproute2 with color support

### LINKED ISSUE #126: zshrc: alias for color output from ip
Currently `zshrc` sets a number of useful aliases, including color output for `ls` and `grep` if available.
I would like an alias for `ip` (from `iptoute2`) to be introduced, it supports the same option `--color=auto` since May 2015 and release 4.1.0.

eworm-de:
Closes #126

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -2676,6 +2676,11 @@ else
     alias l='command ls -l'
 fi
 
+# use ip from iproute2 with color support
+if ip --color=auto addr >/dev/null 2>&1; then
+    alias ip='command ip --color=auto'
+fi
+
 if [[ -r /proc/mdstat ]]; then
     alias mdstat='cat /proc/mdstat'
 fi


mika:
Merged, thanks for the PR! :tada:

eworm-de:
Thinking about this we could optimize by limiting the command to a single interface, like this:

mika:
@eworm-de sorry, I don't understand what you mean, can you please try to reword, provide further details or some proof of concept code? :)

eworm-de:
By default `ip` shows information for *all* interfaces. This can limit the cpu cycles as just one interface is touched, but that interface `lo` has to be available:

```
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -2677,7 +2677,7 @@ else
 fi
 
 # use ip from iproute2 with color support
-if ip --color=auto addr >/dev/null 2>&1; then
+if ip --color=auto addr show dev lo >/dev/null 2>&1; then
     alias ip='command ip --color=auto'
 fi
 
```

mika:
Oh, that's an excellent idea, I'm all for it! If you can come up with a PR I'd happily take that, if you'd prefer me to take care myself just let me know :)

eworm-de:
See #133 :)

PR #128, opened 2021-10-28
zshrc: discard stderr from vcs_info

eworm-de:
Changing directory to a bare git repository results in error messages
being printed:

fatal: this operation must be run in a work tree

This is really annoying, especially when working with git - you do not
know if your command complained or zsh prompt generated it.

So let's just discard stderr from vcs_info.


ft:
I'd prefer not to discard `stderr` here, because it may mask messages that are important to understand why things misbehave.

If this is a problem for a user, they can just use the updated `VCS_INFO_get_data_git` from a private entry in `$fpath`.

eworm-de:
Already updated that for myself. 😊
Closing the pull request...

PR #129, opened 2021-11-05
Ignore hosts set to invalid/loopback addresses for $_etc_hosts 

### LINKED ISSUE #118: grml-zshrc: Ignore setting up _etc_hosts if the file is too large
In situations where the hosts file is used for ad/malware blocking (e.g. utilizing [Steven Black's hosts file](https://github.com/StevenBlack/hosts)), grml-zshrc's generation of `_etc_hosts` can severely impact performance, especially on underpowered systems with low-end CPUs and HDDs, leading to very long loading times for the shell.

It would be nice if grml-zshrc ignored setting up `_etc_hosts` if it was too big (>1MiB maybe?) or allowed for it to be disabled through `.zshrc.pre`.

Once a decision is reached, I can make a simple PR with that

Atrate:
This significantly improves the loading times of the grml zshrc file in
the case a user decides to block ads/tracking system-wide through the
use of a hosts file where tracking domains are blocked as such:
`0.0.0.0 trackingdomain.com`

Closes #118 

(Note: I'm not too familliar with the syntax used when setting up the _etc_hosts variable, I may have made a mistake here)

Signed-off-by: Atrate <Atrate@protonmail.com>

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -894,7 +894,7 @@ function grmlcomp () {
     if is42 ; then
         [[ -r ~/.ssh/config ]] && _ssh_config_hosts=(${${(s: :)${(ps:\t:)${${(@M)${(f)"$(<$HOME/.ssh/config)"}:#Host *}#Host }}}:#*[*?]*}) || _ssh_config_hosts=()
         [[ -r ~/.ssh/known_hosts ]] && _ssh_hosts=(${${${${(f)"$(<$HOME/.ssh/known_hosts)"}:#[\|]*}%%\ *}%%,*}) || _ssh_hosts=()
-        [[ -r /etc/hosts ]] && [[ "$NOETCHOSTS" -eq 0 ]] && : ${(A)_etc_hosts:=${(s: :)${(ps:\t:)${${(f)~~"$(</etc/hosts)"}%%\#*}##[:blank:]#[^[:blank:]]#}}} || _etc_hosts=()
+        [[ -r /etc/hosts ]] && [[ "$NOETCHOSTS" -eq 0 ]] && : ${(A)_etc_hosts:=${(s: :)${(ps:\t:)${${(f)~~"$(grep -v '^0\.0\.0.\0\|^127\.0\.0\.1\|^::1 ' /etc/hosts)"}%%\#*}##[:blank:]#[^[:blank:]]#}}} || _etc_hosts=()
     else
         _ssh_config_hosts=()
         _ssh_hosts=()


mika:
LGTM, I'm only wondering, whether the `0.0.0.0` could be filtered out by zsh syntax, without having to rely on external `grep` invocation? @ft maybe you have some magic™ for us? :smiley:

moviuro:
If we're filtering out `0.0.0.0`, we might as well filter `127.0.0.1` (or maybe even `127/8`?).

mika:
@moviuro yeah, I'm using `127.0.0.1` for this myself, would be nice to have a proper filter list that works for such use cases :)

Atrate:
In such a case it would be nice to filter out `::1` as well, for IPv6. Feel free to push to my forked branch to update the PR with a list or ZSH magic

evgeni:
FWIW, if going with `grep`, this should be `grep -v '^0\.0\.0.\0'` as otherwise you'd skip hosts like 10.0.0.0 (the 1 comes before your regex) or 10.0.010.123 (slightly silly notation, but a valid legacy IP address)

Atrate:
Good catch. Would you like me to update the PR or are we trying for a way not involving `grep`?

ft:
I think that expression is too complex as it is. We should make this into something more readable. Not every zsh expression has to be an unreadable oneliner.

Atrate:
Force-pushed, fixed the expression, added `127.0.0.1` and `::1`.

etc/zsh/zshrc
```diff
@@ -893,7 +893,7 @@ function grmlcomp () {
     if is42 ; then
         [[ -r ~/.ssh/config ]] && _ssh_config_hosts=(${${(s: :)${(ps:\t:)${${(@M)${(f)"$(<$HOME/.ssh/config)"}:#Host *}#Host }}}:#*[*?]*}) || _ssh_config_hosts=()
         [[ -r ~/.ssh/known_hosts ]] && _ssh_hosts=(${${${${(f)"$(<$HOME/.ssh/known_hosts)"}:#[\|]*}%%\ *}%%,*}) || _ssh_hosts=()
-        [[ -r /etc/hosts ]] && : ${(A)_etc_hosts:=${(s: :)${(ps:\t:)${${(f)~~"$(</etc/hosts)"}%%\#*}##[:blank:]#[^[:blank:]]#}}} || _etc_hosts=()
+        [[ -r /etc/hosts ]] && : ${(A)_etc_hosts:=${(s: :)${(ps:\t:)${${(f)~~"$(grep -v '^0\.0\.0.\0\|^127\.0\.0\.1\|^::1 ' /etc/hosts)"}%%\#*}##[:blank:]#[^[:blank:]]#}}} || _etc_hosts=()
```
Atrate, line 896: The space after `::1 ` is so that an address like `::1:2345` isn't included, as it would expand to `0:0:0:0:0:0:0001:2345`

mika:
So what are we going to do with this? (It needs to be rebased against current master for sure :))
Are we fine with this approach for the time being, @evgeni + @ft?

nitpick: the commit message summary line should be "zshrc: ignore hosts set to invalid/loopback addresses for $_etc_hosts" (so start with "zshrc" and also fix the typo in "loopback") :)

evgeni:
I have no real opinion here, just a Zsh user ;)

Atrate:
Rebased on master, force-pushed, @mika.

mika:
Thank you, @Atrate - merged and pushed! :partying_face: 

I noticed that we don't yet handle entries like `::1 example.org` with a tab space as separator character, and also entries like `127.0.1.1 …` aren't ignored yet, we might further want to improve on this, if anyone can take care I'd appreciate that!

moviuro:
I usually use the same bogus IP address for all bad hosts. Maybe we should ignore hosts that have a specific IP address?

```
_ignore_hosts=127.0.1.1
_ignore_hosts=10.10.40.40
...
```

Atrate:
> I usually use the same bogus IP address for all bad hosts. Maybe we should ignore hosts that have a specific IP address?
> 
> ```
> _ignore_hosts=127.0.1.1
> _ignore_hosts=10.10.40.40
> ...
> ```

It may actually be a nice idea to make an array/list of IP addresses to ignore (e.g. 127.0.0.1, etc.) and let the user append to it/clear it if need be. And use that array in place of the grep regexp.

PR #130, opened 2021-11-08
zshrc: add `NOETCHOSTS` parameter

rvolosatovs:
`/etc/hosts` may be used as a denylist for undesired hosts on some systems.
In such usage scenarios the file may be large in size and parsing it may severely impact shell startup time.

For example, on a system with `/etc/hosts` from https://github.com/StevenBlack/hosts:

```zsh
% cat profile.zsh
#!/usr/bin/env zsh
zmodload zsh/zprof
source ./etc/zsh/zshrc
zprof
```

```
% ./profile.zsh | head -3
num  calls                time                       self            name
-----------------------------------------------------------------------------------
 1)    1        1181.93  1181.93   87.91%   1181.14  1181.14   87.85%  grmlcomp
```

```
% NOETCHOSTS=1 ./profile.zsh | head -3
num  calls                time                       self            name
-----------------------------------------------------------------------------------
 1)    2          79.77    39.88   47.37%     79.77    39.88   47.37%  compaudit
```

### DIFF
--- a/doc/grmlzshrc.t2t
+++ b/doc/grmlzshrc.t2t
@@ -100,6 +100,10 @@ A value of zero turns off checking.
 : **NOCOR**
 Non zero values deactivate automatic correction of commands.
 
+: **NOETCHOSTS**
+Non zero values deactivate parsing of "/etc/hosts" disabling host completion
+using file's contents.
+
 : **NOMENU**
 If set to zero (default), allows selection from a menu, if there are at least
 five possible options of completion.
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -481,6 +481,7 @@ setopt unset
 
 # setting some default values
 NOCOR=${NOCOR:-0}
+NOETCHOSTS=${NOETCHOSTS:-0}
 NOMENU=${NOMENU:-0}
 NOPRECMD=${NOPRECMD:-0}
 COMMAND_NOT_FOUND=${COMMAND_NOT_FOUND:-0}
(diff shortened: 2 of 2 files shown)


ft:
I like it. I'd maybe call the parameter `NOETCHOSTS` to be more in line with the other parameters, that don't use underscores.

rvolosatovs:
> I like it. I'd maybe call the parameter `NOETCHOSTS` to be more in line with the other parameters, that don't use underscores.

Done, PTAL!

rvolosatovs:
Oh, didn't realize there's already https://github.com/grml/grml-etc-core/pull/129 tackling the same issue (https://github.com/grml/grml-etc-core/issues/118), albeit in a different way.
I think both approaches could complement each other, in fact.

rvolosatovs:
Oh, didn't realize there's already https://github.com/grml/grml-etc-core/pull/129 tackling the same issue (https://github.com/grml/grml-etc-core/issues/118), albeit in different way
I think both approaches could complement each other, in fact.

mika:
I'm fine with merging this (since skipping `/etc/hosts` independent of its content might be relevant for some users), while we should also find a proper solution for #129 (I'd appreciate your feedback there, @ft!).

We should document this new `NOETCHOSTS` variable in `doc/grmlzshrc.t2t` though, is that something you could also please take care of, @rvolosatov?

Thanks, everyone!

rvolosatovs:
> I'm fine with merging this (since skipping `/etc/hosts` independent of its content might be relevant for some users), while we should also find a proper solution for #129 (I'd appreciate your feedback there, @ft!).
> 
> We should document this new `NOETCHOSTS` variable in `doc/grmlzshrc.t2t` though, is that something you could also please take care of, @rvolosatov?
> 
> Thanks, everyone!

Done, also made sure that alphabetic ordering of parameters is maintained.

rvolosatovs:
@mika any update on this?

mika:
@rvolosatovs  sorry for the delay, we just looked at it and it's perfectly fine as-is, merged! Thanks for your work! :+1:

PR #133, opened 2021-12-03
zshrc: make ip touch interface lo only

eworm-de:
This reduces cpu cycles, but interface `lo` has to be available
for the check to succeed.


mika:
Thanks! :partying_face:

PR #134, opened 2021-12-21
Revert "zshrc: use ip from iproute2 with color support"

pasja:
This reverts commit ae49957322c648fcd64b7e73485cb06f4b694207.

This breaks the upstream zsh completion for the 'ip' command.
In the world where we have interface names like 'enp0s31f6' and
'enxc03ebac7d8c3' I would rather have working auto-complete than some coloring.


jkirk:
You are right. command completion fails with this alias:

```
❯ ip
parse failed before current word
```

I played around a bit and found we can have both (command completion with color support) when changing the alias to `ip -color=auto` (notice the single dash).
Could you try that and please update the PR accordingly? Thx.

pasja:
Thanks for the investigation. I have tried it on my machine and can confirm you results.
I have updated the PR with your fix.

moviuro:
Wouldn't overloading `ip` with a function instead of an alias work?

```
moviuro@toxoplasmosis ~ % which ip
ip () {
        command ip --color=auto "$@"
}
moviuro@toxoplasmosis ~ % ip
completing ip command
addr       -- manage protocol address
addrlabel  -- manage addrlabel
link       -- configure network device
maddr      -- manage multicast addresses
monitor    -- monitor state
mroute     -- manage multicast routing cache
neigh      -- manage neighbour/ARP tables
route      -- manage routing table
rule       -- manage routing policy database
tunnel     -- configure tunnel
```

mika:
Interesting find! If the usage of a function, instead of an alias, fixes the issue, I'd prefer this approach.

But furthermore, regarding the completion handling itself: `/usr/share/zsh/functions/Completion/Unix/_ip` uses `$(ip addr show | sed -n ...`, shouldn't this be `$(command ip addr show | sed -n ...`, or even better `$(_call_program commands ip addr show | sed -n ...`)`? @ft what do you think, should we raise this upstream?

jkirk:
The ip(8) man page only and the ip help only mention single dash options, the double-dash options only seem to work "by chance" (but yes, command completion should also work with double dash options)... 

Any reason why we should prefer functions over aliases? (I'd prefer the simple alias in this case, so the PR lgtm.)

pasja:
@mika can I help you with anything regarding this PR?

mika:
@pasja nope, I'd say let's take your fix as-is and if someone has any further suggestions for improving it, we can handle this through a separate PR. Thanks for your PR!

PR #138, opened 2022-03-04
tmux.conf: Add bind-key J to join marked pane to active window

jkirk:
We have "bind-key j" which joins "the last active pane to the currently
active window". This is cool, if you are able to directly jump to the
window where the pane should go to. Old people like me do not remember
the window (number) where they came from and skim around until they find
the window in question, try to remember the window number, jump back and
forth from window with the needed pane and finally "join the last active
pane to the currently active window".

This is a bit cumbersome, but with "bind-key m" it is possible to
mark the pane in question, skim around until one fine the window where
the pane belongs and "join the marked pane to the currently active
window".

### DIFF
--- a/etc/tmux.conf
+++ b/etc/tmux.conf
@@ -22,6 +22,10 @@ bind-key '\' confirm-before kill-session
 ### join the last active pane to the currently active window
 bind-key j join-pane -s !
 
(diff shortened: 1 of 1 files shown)


mika:
LGTM, thanks!

PR #142, opened 2022-07-25
zshrc: smarter smarter cd

moviuro:
We have to account for a lot of possibilities (see zshbuiltins(1)): we stay on the safe side and only change cd target if we're sure about what we do. See examples:

```
% zsh -d -f             # reset the shell
% cd
% touch -- -q
% cd -q                 # should do nothing; 43090cf version behaves weirdly here, replacing -q with .
% cd /home
% cd /home /etc         # should go to /etc (replace home with etc in PWD)
/etc
% cd /home /etc         # this will fail [1]
cd: string not in pwd: /home
% cd /home
% cd -q /home /etc      # should go to /etc
/etc
% cd -q /home /etc      # this will fail [2]
cd: string not in pwd: /home
% cd
% touch +123456
% cd +123456            # this will fail [3]; 43090cf version behaves weirdly here, replacing +123456 with .

[1]   I chose to make this one work. If `cd old new` fails, then try to
      `cd old`
[2,3] I chose to fail too in that case. Handling cd arguments is not what
      I was planning to do when I started working on this patch ;P
```

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -3470,14 +3470,27 @@ function cl () {
     cd $1 && ls -a
 }
 
-# smart cd function, allows switching to /etc when running 'cd /etc/fstab'
+# smart cd function, allows switching to /etc when running 'cd /etc/fstab',
+# `cd /etc/fst` or `cd /etc/*tab`
 function cd () {
-    if (( ${#argv} == 1 )) && [[ -f ${1} ]]; then
-        [[ ! -e ${1:h} ]] && return 1
-        print "Correcting ${1} to ${1:h}"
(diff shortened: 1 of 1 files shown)


mika:
Thanks, @moviuro - any objections against merging this as-is from your side, @ft?

moviuro:
You *probably* want to run some more tests on it. I ran a few and made sure my usecases were covered, though I might have missed some.

Even got to learn about zshbuiltins(1) in the process!

ImOdd (approved):
Word good look n

mika:
@ft maybe you could please take a look at this once you've some spare minutes? :pray:

PR #146, opened 2022-09-05
zshrc: fix the conditional statement

miles170:
fix bug introduced by https://github.com/grml/grml-etc-core/commit/220d6ffaba8a58e71c6c51d30196c6fe07f18e37


mika:
Hm, what exactly did I break with my change in https://github.com/grml/grml-etc-core/commit/220d6ffaba8a58e71c6c51d30196c6fe07f18e37? AFAICS, this change here modifies behavior such that `~/.ssh/known_hosts` and `/etc/hosts` are no longer supported then with zsh >=4.2, or am I missing something? :thinking:

mika:
Sigh, looked at it three times and didn't spot it :( @miles170 is https://github.com/grml/grml-etc-core/pull/147 what you had in mind?

miles170:
> Sigh, looked at it three times and didn't spot it :( @miles170 is #147 what you had in mind?

Yes, thanks!

mika:
Great, thanks for reporting in such a timely fashion! :)

PR #149, opened 2022-10-03
zshrc: run tlog alias with --follow=name

mika:
It's annoying to follow the syslog, but then tools like logroate kicking in to rename syslog to syslog.1. Then you're stuck with the "old" syslog file content, by using the --follow=name option we can fix this.

Quoting from tail(1):

```
With --follow (-f), tail defaults to following the file descriptor, which means that even
if a tail'ed file is renamed, tail will continue | to track its end. This default behavior
is not desirable when you really want to track the actual name of the file, not the
file descriptor (e.g., log rotation). Use --follow=name in that case. That causes tail
to track the named file in a way that accommodates renaming, removal and creation.
```

### DIFF
--- a/doc/grmlzshrc.t2t
+++ b/doc/grmlzshrc.t2t
@@ -1106,7 +1106,7 @@ http://www.cl.cam.ac.uk/~mgk25/unicode.html#term).
 Sets mode from ISO 2022 to UTF-8 (See:
 http://www.cl.cam.ac.uk/~mgk25/unicode.html#term).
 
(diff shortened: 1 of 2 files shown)


evgeni (approved):
Makes sense

jkirk (approved):

ft (approved):
Didn't know this option existed. Sounds about right.

mika:
Thanks everyone for your fast feedback! :partying_face: :kissing_heart:

PR #151, opened 2022-11-25
tmux.conf: detect running as tmate and disable non-supported features

mika:
tmux versions >=2.3 support the -q option for source-file to suppress errors for nonexistent files. Even Debian/oldoldstable has tmux v2.3-4, so this works nice everwhere, *except* for when invoking tmate, which is a fork based on an old version of tmux, which then complains about:

| /etc/tmux.conf:75: usage: source-file path
| /etc/tmux.conf:36: usage: source-file path

Try to detect tmate from within the environment, and then enable the configuration only when *not* running from within tmate.

### DIFF
--- a/etc/tmux.conf
+++ b/etc/tmux.conf
@@ -30,7 +30,8 @@ bind-key J join-pane
 bind-key B if-shell "! tmux has-session -t bg" "new-session -d -s bg" \; move-window -t bg
 
 ### Reload Config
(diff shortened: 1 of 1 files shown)


mika:
@jkirk can you please check whether this works as expected? (I'm really not sure, as the `bindkey R` doesn't seem to behave as expected for me, or I'm holding it wrong :-/)

mika:
Superseded by https://github.com/grml/grml-etc-core/pull/152, thanks @jkirk!

PR #158, opened 2024-10-02
Stop shipping /etc/zsh/zlogout to avoid clearing terminal screen

mika:
/etc/zsh/zlogout causes the clearing of the terminal screen. It was introduced in commit c31db6c8f535a back in 2006, though several of us Grml developers actually don't like this behavior.    

While working on this, had to touch dpkg-divert, again, and had to dig deeper, quoting from the resulting commit:

```
debian/preinst: drop deprecated dpkg-divert
    
Usage of backup_conffile was introduced for moving diverted configuration
files from package grml-etc to grml-etc-core. We seem to have migrated
this in 2006, so there really shouldn't any users out there anymore. :)
  
Also the file /etc/skel/.zshrc is no longer shipped with Debian's zsh
package since v4.3.0-dev-2-1, which dates back to 2011-02-11 (see git
rev 78ab02f in https://salsa.debian.org/debian/zsh.git). Even oldoldstable
has v5.7.1, so finally drop this workaround.

So take this as chance to clean up debian/preinst at least a bit.
```
So take this as chance to clean up debian/preinst at least a bit.


zeha (approved):

zeha:
lgtm

jkirk (approved):

PR #159, opened 2024-10-07
tmux: status-right: Show IP addresses (similar to our GNU/screen configuration)

jkirk:


etc/tmux.conf
```diff
@@ -59,8 +59,11 @@ set -g status-left "#[fg=white]@#h#[fg=red]:#S#[fg=white] |"
 #set -g status-left "#[fg=blue]#(echo ${USER})#[fg=white]@#h#[fg=red]:#S#[fg=white] |"
 
 ### status-right: Date and Time
-set -g status-right-length 16
-set -g status-right '#[fg=yellow]%Y-%m-%d %H:%M'
+#set -g status-right-length 16
+#set -g status-right '#[fg=yellow]%Y-%m-%d %H:%M'
+### status-right: IPs and Date and Time
+set -g status-right-length 100
+set -g status-right "#[fg=green](net: #(hostname -I | sed -e 's/ [^ ]*:.*$//;s/ / | /g;s/ | $//')) #[fg=yellow]%Y-%m-%d %H:%M"
```
moviuro, line 66: FWIW, `hostname -I` is not portable to e.g. OpenBSD: https://man.openbsd.org/hostname.1

mika: > FWIW, `hostname -I` is not portable to e.g. OpenBSD: https://man.openbsd.org/hostname.1

Excellent point, thanks @moviuro! I think we should go through our wrapper script then, which only runs `hostname -I` if that's available? (And possibly fall back to something else otherwise, if at all?)

zeha: if we do wrapper scripts we can stick to the perl variant

zeha: (its not like ip-screen worked on openbsd either)

mika: > if we do wrapper scripts we can stick to the perl variant

wondering whether it would make sense to have `ip-screen` inside $PATH which then is simple script to check for `hostname -I` support and if not supported falls back to your perl script otherwise? :grin:

zeha: Maybe it's worth having separate config file sets, one to use on the live CD, and one for people pulling the individual files. Not sure how f.e. screenrc worked before for individual-config-file users?

zeha (changes requested):
@jkirk will come with a _generic version etc

etc/tmux.conf
jkirk: I played around with a "if-shell" line to check if `hostname` exists. I'll update this PR.

jkirk: Argl, we are talking about `hostname -I`.

Can we have a decision on this? Keep this `hostname -I` line, implement a wrapper script (again) or drop this PR?

jkirk: FWIW, I came up with a `hostname -I` command line option check. Please (re)review.

mika (approved):
Untested, LGTM

etc/tmux.conf
```diff
@@ -59,8 +59,11 @@ set -g status-left "#[fg=white]@#h#[fg=red]:#S#[fg=white] |"
 #set -g status-left "#[fg=blue]#(echo ${USER})#[fg=white]@#h#[fg=red]:#S#[fg=white] |"
 
 ### status-right: Date and Time
-set -g status-right-length 16
-set -g status-right '#[fg=yellow]%Y-%m-%d %H:%M'
+#set -g status-right-length 16
+#set -g status-right '#[fg=yellow]%Y-%m-%d %H:%M'
+### status-right: IPs and Date and Time
+set -g status-right-length 100
```
zeha, line 65: what exactly does this change and is this a good idea on 80x25 screens?

jkirk: Hmm. Good question. It was previously set to 16 which was too short, I set it arbitrary to 100.

I just checked to man page, the default is 40, which should be fine for at least one IP:
```
❯ echo "(net: 192.168.123.254) 2025-02-24 09:43" | wc -c
40
```

FTR, my right status line (`(net: 10.17.0.209 | 10.10.2.100 | 10.17.0.128) 2025-02-24 09:48`) is also shown fine without setting ` status-right-length`. I'll drop it the line in the PR.

jkirk: Updated the PR. Thx!

mika (approved):

Issue #160, opened 2024-10-13
grml-zsh-config does not check if history related variables have been set before setting them


unknowndevQwQ:
https://github.com/grml/grml-etc-core/blob/bdc51f9788c98839cfcb248b9a16d87c8f99c3e2/etc/zsh/zshrc#L1683-L1685
I switched to grml's zshrc and used it for quite some time before noticing that the history size was much smaller than what was defined in /etc/profile.d. I'm going to submit a PR for this, but I'm not sure which of the three ways to change the HISTSIZE and SAVEHIST variables I should choose.

```
[[ -z "${HISTFILE}" ]] && HISTFILE=${HISTFILE:-${ZDOTDIR:-${HOME}}/.zsh_history} 
```
1.
```
if [[ -z "${HISTSIZE}" ]]; then
  isgrmlcd && HISTSIZE=500  || HISTSIZE=5000
fi
if [[ -z "${SAVEHIST}" ]]; then
  isgrmlcd && SAVEHIST=1000 || SAVEHIST=10000 # useful for setopt append_history 
fi
```
2.
```
[[ -z "${HISTSIZE}" ]] && { isgrmlcd && HISTSIZE=500  || HISTSIZE=5000; }
[[ -z "${SAVEHIST}" ]] && { isgrmlcd && SAVEHIST=1000 || SAVEHIST=10000; } # useful for setopt append_history 
```
3.
```
isgrmlcd && HISTSIZE=500  || HISTSIZE=${HISTSIZE:-5000}
isgrmlcd && SAVEHIST=1000 || SAVEHIST=${SAVEHIST:-10000} # useful for setopt append_history 
```

unknowndevQwQ:
> Have you tried adding HISTSIZE to `.zshrc.local`? This is where I did set the variable to override the default.

Setting it again in .zshrc.local solves the problem, but I have the same history length set for every shell in /etc/profile.d and don't want to set it again for zsh.

mika:
I agree, let's go for option 1) though I'd say. :)

ft:
FWIW, I'd go for the `foo=${foo:-default}` option.

mika:
FTR, we need to find another approach, this breaks user setups, see https://github.com/grml/grml-etc-core/issues/169

PR #163, opened 2024-12-06
fix(zshrc): keep existing history restrictions

unknowndevQwQ:
fix: https://github.com/grml/grml-etc-core/issues/160


mika (approved):

mika:
Thank you, @unknowndevQwQ!

PR #166, opened 2024-12-10
zshrc: increase HISTSIZE + SAVEHIST to our defaults also for live system

mika:
Otherwise it might be actually unexpected for some users of the Grml live system, so use same defaults for live system as for non-live systems.

While at it, switch to the foo=${foo:-default} style as suggested by Frank Terbeck in https://github.com/grml/grml-etc-core/issues/160 - thanks!

FTR: maybe we want to even further increase the defaults? (Personally I've set it to `10000000` on my own systems, not sure it's relevant memory wise or so?)


zeha:
s/stile/style/ in commit message + PR text

zeha:
I'll merge this now, but any feedback would still be appreciated!

PR #167, opened 2024-12-11
debian: use modern build system

zeha:
- Switch debian/rules to debhelper sequencer, and use install, manpages files to install the actual files.
- Rename (and update) lintian overrides so debhelper finds them.
- Set debian/source/format to 3.0 (native), as lintian really wants the file.

Addresses grml/grml#114


mika (approved):
Thanks! :heart:

Issue #169, opened 2024-12-17
History not saved since 0.19.9 when SAVEHIST is unset


Derppening:
After upgrading to 0.19.9, I realize that none of my history is saved anymore. 

I tried to debug this issue using `zsh -x` and noticed this line:

```
+/etc/zsh/zshrc:1684> HISTSIZE=30 
+/etc/zsh/zshrc:1685> SAVEHIST=0
```

I think with the change made in #166 didn't take into account that Zsh sets these variables by default, as shown below:

```
% zsh -d -f -i
% echo $HISTSIZE
30
% echo $SAVEHIST
0
```

Thus the `:-5000` and `:-10000` defaults that were present in the change is never applied.

I understand that I can manually set `HISTSIZE` and `SAVEHIST` in `$HOME/.zshrc`, but this behavior may be surprising to some users as the defaults are changed with no indication (e.g. a minor version bump).

mika:
@Derppening uh nice catch, thanks for reporting this! We'll definitely need to address this, will take care ASAP

unknowndevQwQ:
> I've uploaded grml-etc-core v0.19.10 which reverts the changes from #160 for now. Not sure whether we have a good way how to handle this, given that Zsh upstream sets `HISTSIZE=30` and `SAVEHIST=0` by default. :-/
> 
> @ft any idea what we could do about this?

```
[[ -z "${HISTSIZE}" ]] || (("${HISTSIZE}" >  30))
[[ -z "${SAVESIZE}" ]] || (("${SAVESIZE}" >  0))
```
I think...

zeha:
```sh
if [ "${HISTSIZE}" = "30" ]; then
  # override settings only if defaults from zsh are in place
  HISTSIZE=5000
  SAVEHIST=10000 # useful for setopt append_history
fi
```

?

mika:
Yeah, checking for HISTSIZE might be worth an idea! I'm only afraid that this might break as soon as zsh upstream changes this:

```
config.h.in:#define DEFAULT_HISTSIZE 30
configure.ac:#define DEFAULT_HISTSIZE 30
```

:thinking:

zeha:
> Yeah, checking for HISTSIZE might be worth an idea! I'm only afraid that this might break as soon as zsh upstream changes this...

Well they haven't changed it since 1999 at least (git commit "Initial revision"). If they ever change it, grml zshrc users will need an update.

unknowndevQwQ:
[...]

PR #170, opened 2024-12-18
zshrc: set HISTSIZE + SAVEHIST only if not yet set or being set to zsh upstream default

mika:
In commit 7465ffb0803b0eb05d808b850088e90572c8dacd we adjusted the behavior when HISTSIZE and/or SAVEHIST are unset. Though this broke existing setups where HISTSIZE and SAVEHIST was not yet set (like in /etc/profile.d as reported in #160), as zsh sets HISTSIZE and SAVEHIST by default:

```
  config.h.in:#define DEFAULT_HISTSIZE 30
  configure.ac:#define DEFAULT_HISTSIZE 30
```

So `zsh -f` defaults to $HISTSIZE of 30 and $SAVEHIST of 0, Now we clearly want to ship better defaults than this. Do this, by checking whether is HISTSIZE is unset or set to the upstream default (being 30), only then set HISTSIZE and SAVEHIST.

In commit 81096c8f57c088163cfcea49f2eb03b4b44c66fd we increased HISTSIZE + SAVEHIST also for our live system usage, re-apply this change.

Closes: https://github.com/grml/grml-etc-core/issues/160 and https://github.com/grml/grml-etc-core/issues/169

### DIFF
--- a/doc/grmlzshrc.t2t
+++ b/doc/grmlzshrc.t2t
@@ -90,8 +90,7 @@ grml-small.
 Where zsh saves the history. Default: ${HOME}/.zsh_history.
 
 : **HISTSIZE**
-Number of commands to be kept in the history. On a Grml-CD this defaults to
-500, on a hard disk installation to 5000.
+Number of commands to be kept in the history. Default: 5000
 
 : **MAILCHECK**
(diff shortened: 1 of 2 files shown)


mika:
FYI @unknowndevQwQ + @Derppening

PR #172, opened 2024-12-21
zshenv: stop setting HOSTNAME

zeha:
In a normal zsh environment this variable is called HOST, and is automatically set.

### DIFF
--- a/etc/zsh/zshenv
+++ b/etc/zsh/zshenv
@@ -25,13 +25,6 @@
 # no xsource() here because it's only created in zshrc! (which is good)
 [[ -r /etc/environment ]] && source /etc/environment
 
-# set environment variables (important for autologin on tty)
-if [ -n "${HOSTNAME}" ] ; then
-    export HOSTNAME="${HOSTNAME}"
-else
-    export HOSTNAME="$(uname -n)"
-fi
-
 # make sure /usr/bin/id is available
 if [[ -x /usr/bin/id ]] ; then
     [[ -z "$USER" ]]          && export USER=$(/usr/bin/id -un)


zeha:
$HOSTNAME might be relevant for `grml_maintain_name` in zshrc, but tbh I cannot figure out how it is meant to work in the first place. Maybe it should also go.

mika:
@ft maybe you could assist us with the `grml_maintain_name` stuff in grml-zshrc? :) AFAICT it was initially added to set the host name if not running on a host with name `grml`. Over time it was adapted furthermore, the `NAME="@$HOSTNAME"` and the lines below the `grml_cmd_to_screen_title` function feels like it's about "workarounds" for GNU screen?

Where is $HOSTNAME actually supposed to be coming from? Should this be set by our zsh at all?

ft:
So… There is some old stuff involved here, some of the lines get blamed to `"initial checkin"`. :)

As an aside, us using `$NAME` here is a little dirty because that seems like a parameter name, other people may override, for whatever reason. We also do have a lot of configurability in here, that makes the code a little hard to follow. Like, do we really need things like `$NOPRECMD` and `$NOTITLE`? We also have a copy of `add-zsh-hook` in our code base. We did that, to work around a bug with old zsh versions. The comment and commit-message say we can remove this, whenever we update the minimum zsh version to `4.3.7`. That was released in 2008. I think it's time to do this, so we get the newest version of this function from upstream, via autoload (which we have to re-enable in line 2632).

The functions in this area (`grml_reset_screen_title`, `grml_vcs_to_screen_title`, `grml_maintain_name`, `grml_cmd_to_screen_title`, and `grml_control_xterm_title`), are registered to hooks. These hooks are used to make things like the `precmd` and `preexec` functions more easily extensible (basically like hooks work in emacs, too, if you've seen that).

`precmd` is called just before a prompt is drawn (and redrawn for instance when a job completes, or when fancy completion actions need a prompt redraw). `preexec` is called just after you press enter and a command line is about to be executed. These points in time make them useful for the updating things in your prompt (vcs-info is called in `precmd`, for instance).

The additional functions to work on things like the prompt, namely the terminal emulator titles and window titles in things like GNU screen (maybe tmux is also supported, I don't recall). So it's not really a workaround for screen. Rather it's a question of whether or not you want to update the title of a GNU screen window. Right now what should happen is this:

- While zsh's command prompt is waiting for enter: Put version control information in there.
- While a command is executing: Put the command name that executes in there.

Here's a patch to remove the `add-zsh-hook` function definition:

```udiff
diff --git a/etc/zsh/zshrc b/etc/zsh/zshrc
index e6815e4..799926e 100644
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -338,82 +338,8 @@ function zrcautoload () {
     return 0
 }
 
-# The following is the ‘add-zsh-hook’ function from zsh upstream. It is
-# included here to make the setup work with older versions of zsh (prior to
-# 4.3.7) in which this function had a bug that triggers annoying errors during
-# shell startup. This is exactly upstreams code from f0068edb4888a4d8fe94def,
-# with just a few adjustments in coding style to make the function look more
-# compact. This definition can be removed as soon as we raise the minimum
-# version requirement to 4.3.7 or newer.
-function add-zsh-hook () {
-    # Add to HOOK the given FUNCTION.
-    # HOOK is one of chpwd, precmd, preexec, periodic, zshaddhistory,
-    # zshexit, zsh_directory_name (the _functions subscript is not required).
-    #
-    # With -d, remove the function from the hook instead; delete the hook
-    # variable if it is empty.
-    #
-    # -D behaves like -d, but pattern characters are active in the function
-    # name, so any matching function will be deleted from the hook.
-    #
-    # Without -d, the FUNCTION is marked for autoload; -U is passed down to
-    # autoload if that is given, as are -z and -k. (This is harmless if the
-    # function is actually defined inline.)
-    emulate -L zsh
-    local -a hooktypes
-    hooktypes=(
-        chpwd precmd preexec periodic zshaddhistory zshexit
-        zsh_directory_name
-    )
-    local usage="Usage: $0 hook function\nValid hooks are:\n  $hooktypes"
-    local opt
-    local -a autoopts
-    integer del list help
-    while getopts "dDhLUzk" opt; do
-        case $opt in
-        (d) del=1 ;;
-        (D) del=2 ;;
-        (h) help=1 ;;
-        (L) list=1 ;;
-        ([Uzk]) autoopts+=(-$opt) ;;
-        (*) return 1 ;;
-        esac
-    done
-    shift $(( OPTIND - 1 ))
-    if (( list )); then
-        typeset -mp "(${1:-${(@j:|:)hooktypes}})_functions"
-        return $?
-    elif (( help || $# != 2 || ${hooktypes[(I)$1]} == 0 )); then
-        print -u$(( 2 - help )) $usage
-        return $(( 1 - help ))
-    fi
-    local hook="${1}_functions"
-    local fn="$2"
-    if (( del )); then
-        # delete, if hook is set
-        if (( ${(P)+hook} )); then
-            if (( del == 2 )); then
-                set -A $hook ${(P)hook:#${~fn}}
-            else
-                set -A $hook ${(P)hook:#$fn}
-            fi
-            # unset if no remaining entries --- this can give better
-            # performance in some cases
-            if (( ! ${(P)#hook} )); then
-                unset $hook
-            fi
-        fi
-    else
-        if (( ${(P)+hook} )); then
-            if (( ${${(P)hook}[(I)$fn]} == 0 )); then
-                set -A $hook ${(P)hook} $fn
-            fi
-        else
-            set -A $hook $fn
-        fi
-        autoload $autoopts -- $fn
-    fi
-}
+# Autoload zsh's hook system early, so we can use it whenever we need.
+zrcautoload add-zsh-hook || add-zsh-hook () { :; }
 
 # Load is-at-least() for more precise version checks Note that this test will
 # *always* fail, if the is-at-least function could not be marked for
@@ -2626,10 +2552,6 @@ function grml_control_xterm_title () {
     esac
 }
 
-# The following autoload is disabled for now, since this setup includes a
-# static version of the ‘add-zsh-hook’ function above. It needs to be
-# re-enabled as soon as that static definition is removed again.
-#zrcautoload add-zsh-hook || add-zsh-hook () { :; }
 if [[ $NOPRECMD -eq 0 ]]; then
     add-zsh-hook precmd grml_reset_screen_title
     add-zsh-hook precmd grml_vcs_to_screen_title
```

Hope that helps. :)

ft:
There is also lots of stuff like

```sh
if is433 ; then
```

…that could be simplified if we updated the minimum version requirement.

FWIW, Debian `oldstable` has zsh 5.8, which was released in 2020. The most recent compatibility check that we have is `is51` which checks for zsh `5.1.0` and newer. That thing was released in 2015. So if we update the minimum requirement to that, we can get rid of all that compatibility stuff. In the past usually centos was the most widely used system with the most ancient version of zsh. Not sure how to find out what they use these days. But if they are behind by a decade, I wouldn't even bother.

ft:
Oh, and yeah, the old code just made sure that there is a `HOSTNAME` parameter in the environment. POSIX doesn't list it as one of the common environment parameters, that you should stay away from¹. My private zsh setup uses `$HOST` in one place, which is indeed set in my environment setup, without me setting it. But I think it's probably best to use `$(hostname)` instead, for the local hostname. It looks to me like we use `HOSTNAME`, just to be able to tell if we're in a remote connection. Not sure if that's wise.

In any case, this:

```sh
function grml_maintain_name () {
    local localname
    localname="$(uname -n)"

    # set hostname if not running on local machine
    if [[ -n "$HOSTNAME" ]] && [[ "$HOSTNAME" != "${localname}" ]] ; then
       NAME="@$HOSTNAME"
    fi
}

function grml_cmd_to_screen_title () {
    # get the name of the program currently running and hostname of local
    # machine set screen window title if running in a screen
    if [[ "$TERM" == screen* ]] ; then
        local CMD="${1[(wr)^(*=*|sudo|ssh|-*)]}$NAME"
        ESC_print ${CMD}
    fi
}
```

Seems to just use `$NAME` to inject the original hostname in the pattern to strip down the command name used in the screen title. I think this is probably to strip down things like `ssh user@hostname command here …` a little more.

If we don't set `$HOSTNAME` any more, then the `$NAME` stuff also won't work (unless something else sets `$HOSTNAME` out of our control). So that should go as well.

My private setup uses `SSH_CLIENT` to do some conditional stuff in zsh sessions that run on a host via ssh.  But it seems that variable has since been at least deprecated. Not sure if that even still works.

¹ https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html

ft:
Okay, final remark (maybe): `HOSTNAME` is a variable (but **not** an environment variable — i.e. it is **not** exported to child processes) in `bash`. Maybe we initially wanted to emulate this `bash` variable, and then someone used it for the `$NAME` magic.

zeha:
> It looks to me like we use HOSTNAME, just to be able to tell if we're in a remote connection. Not sure if that's wise.

This is the part that I really don't understand how this could work. If we assume that most "remote connections" are really SSH, they will never have a differing HOSTNAME.

At least not until someone also modifies their own ssh_config to forward that variable. grml-etc-core does not ship an appropriate ssh_config.

So I don't see how this could actually work (by default) anywhere. IMO we should remove it, do people agree?

zeha:
> In the past usually centos was the most widely used system with the most ancient version of zsh. Not sure how to find out what they use these days. 

@evgeni maybe you could add zsh to canidepend? :)

zeha:
> > In the past usually centos was the most widely used system with the most ancient version of zsh. Not sure how to find out what they use these days.
> 
> @evgeni maybe you could add zsh to canidepend? :)

https://mirror.stream.centos.org/10-stream/BaseOS/aarch64/os/Packages/

has zsh 5.9.x.

evgeni:
> > In the past usually centos was the most widely used system with the most ancient version of zsh. Not sure how to find out what they use these days. 
> 
> @evgeni maybe you could add zsh to canidepend? :)

So can you!

https://github.com/canidepend/canidepend.github.com/blob/master/data/data.yml

ft:
> > It looks to me like we use HOSTNAME, just to be able to tell if we're in a remote connection. Not sure if that's wise.
> 
> This is the part that I really don't understand how this could work. If we assume that most "remote connections" are really SSH, they will never have a differing HOSTNAME.
> 
> At least not until someone also modifies their own ssh_config to forward that variable. grml-etc-core does not ship an appropriate ssh_config.
> 
> So I don't see how this could actually work (by default) anywhere. IMO we should remove it, do people agree?

Yeah, my best guess is, that previous ssh installations maybe didn't whitelist some but only blacklist some variables from the environment. All I can say, that I have never seen this working myself, neither now, nor in the past, due to the fact, that when I'm not using grml, I'm on my own zshrc. :)

I think what we'd want now is this:

```udiff
diff --git a/etc/zsh/zshrc b/etc/zsh/zshrc
index e6815e4..43cedfd 100644
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -2599,21 +2599,11 @@ function grml_vcs_to_screen_title () {
     fi
 }
 
-function grml_maintain_name () {
-    local localname
-    localname="$(uname -n)"
-
-    # set hostname if not running on local machine
-    if [[ -n "$HOSTNAME" ]] && [[ "$HOSTNAME" != "${localname}" ]] ; then
-       NAME="@$HOSTNAME"
-    fi
-}
-
 function grml_cmd_to_screen_title () {
     # get the name of the program currently running and hostname of local
     # machine set screen window title if running in a screen
     if [[ "$TERM" == screen* ]] ; then
-        local CMD="${1[(wr)^(*=*|sudo|ssh|-*)]}$NAME"
+        local CMD="${1[(wr)^(*=*|sudo|ssh|-*)]}"
         ESC_print ${CMD}
     fi
 }
@@ -2633,7 +2623,6 @@ function grml_control_xterm_title () {
 if [[ $NOPRECMD -eq 0 ]]; then
     add-zsh-hook precmd grml_reset_screen_title
     add-zsh-hook precmd grml_vcs_to_screen_title
-    add-zsh-hook preexec grml_maintain_name
     add-zsh-hook preexec grml_cmd_to_screen_title
     if [[ $NOTITLE -eq 0 ]]; then
         add-zsh-hook preexec grml_control_xterm_title
```

But I can't test that stuff right now.

I think the `"${1[(wr)^(*=*|sudo|ssh|-*)]}"` is cargo-culted from somewhere, but I can't find the reference at the moment. my google-fu is failing me. I suppose we could do something more readable and more flexible about that thing. Also, I think there are not `tmux` and `tmux-26color` variants for `TERM`, which we could support as well. Maybe with something like this:

```sh
function inside_multiplexer () {
    case "$TERM" in
    screen|screen-*) return 0 ;;
    tmux|tmux-*)     return 0 ;;
    esac
    return 1
}
```

Tests for `TERM` being something like screen is in a few places in our `zshrc`.

zeha:
Right; I think removing `grml_maintain_name` is a good outcome for now.

My tmux seems to ignore title changes, but anyway we can do that in a separate PR. I like the `inside_multiplexer` idea.

mika (approved):
Thanks, this was more than overdue (all blame on myself), many thanks @zeha @ft + @evgeni for looking into this! :bow:

zeha:
> > > In the past usually centos was the most widely used system with the most ancient version of zsh. Not sure how to find out what they use these days.
> > 
> > 
> > @evgeni maybe you could add zsh to canidepend? :)
> 
> So can you!
> 
> https://github.com/canidepend/canidepend.github.com/blob/master/data/data.yml

@evgeni thanks for https://github.com/canidepend/canidepend.github.com/commit/9c4e159142f96e27eec0badb1f2c7fced42add37 :)

PR #173, opened 2024-12-22
Raise minimum zsh version to 5.1

zeha:
Per discussion in #172

### DIFF
--- a/doc/grmlzshrc.t2t
+++ b/doc/grmlzshrc.t2t
@@ -20,7 +20,7 @@ as //grmlzshrc// throughout this document) for the amazing unix shell zsh
 (http://zsh.sourceforge.net). This is the reference manual for that
 setup.
 
-To use //grmlzshrc//, you need at least version 5.0 of zsh (although not all
+To use //grmlzshrc//, you need at least version 5.1 of zsh (although not all
 features are enabled in every version).
 
 //grmlzshrc// behaves differently depending on which user loads it. For the
--- a/doc/grmlzshrc.t2t
+++ b/doc/grmlzshrc.t2t
@@ -20,8 +20,7 @@ as //grmlzshrc// throughout this document) for the amazing unix shell zsh
 (http://zsh.sourceforge.net). This is the reference manual for that
 setup.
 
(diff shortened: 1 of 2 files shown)


zeha:
@ft I took your patch for removing the zsh-add-hook copy and committed it here. I've also removed a separate commented-out autoload of zsh-add-hook.

Could I bother you into reviewing this?

zeha:
I guess `grmlcomp` could stop being a function, but not sure if this is a good idea.

ft (approved):
I looked through these, and everything looks fine to me. You did the version bump much more careful than I would have. Nice. ;)

ft:
> I guess `grmlcomp` could stop being a function, but not sure if this is a good idea.

Yes, I agree. There is one instance of a `local`'d variable:


```sh
    local localname
    localname="$(uname -n)"
    hosts=(
        "${localname}"
        "$_ssh_config_hosts[@]"
        "$_ssh_hosts[@]"
        "$_etc_hosts[@]"
        localhost
    )
```

But you don't need that variable, if you just use `$(uname -n)` when setting `$hosts` directly.

ft:
Sorry for all the drive-bys, but we may also want:

```udiff
diff --git a/doc/grmlzshrc.t2t b/doc/grmlzshrc.t2t
index 49db5a3..e23eaec 100644
--- a/doc/grmlzshrc.t2t
+++ b/doc/grmlzshrc.t2t
@@ -20,8 +20,7 @@ as //grmlzshrc// throughout this document) for the amazing unix shell zsh
 (http://zsh.sourceforge.net). This is the reference manual for that
 setup.

-To use //grmlzshrc//, you need at least version 3.1.7 of zsh (although not all
-features are enabled in every version).
+To use //grmlzshrc//, you need at least version 5.1 of zsh.

 //grmlzshrc// behaves differently depending on which user loads it. For the
 root user (**EUID** == 0) only a subset of features is loaded by default. This
```

zeha:
> Sorry for all the drive-bys, but we may also want:

No worries, I have added your patch.

zeha:
I'll look more closely at `grmlcomp` in a new PR.

PR #174, opened 2024-12-22
Drop outdated and/or broken things

zeha:
Invididual commits have explanations.

PR #175, opened 2024-12-22
zshrc: use path_helper on macOS

zeha:
macOS provides a system-wide PATH configuration. For that to work, login shells need to evaluate `path_helper -s`. Add this.


zeha:
The fink removal commit is in here, as it would otherwise just conflict after merging #174

mika (approved):
Nice, TIL!

PR #176, opened 2024-12-23
zshrc: unify terminal multiplex detection

zeha:


zeha:
TODO:
- [ ] make sure this actually works in tmux
- [ ] docs?
- [ ] ?

Issue #180, opened 2025-01-16
Remove "working directory" hacks in zshenv/zshrc


jkirk:
@zeha mentioned in https://github.com/grml/grml-live/pull/247 some zshenv/zshrc hacks, which should be removed.

Investigate the hacks and remove them.

PR #181, opened 2025-01-16
zshenv: drop live-cd only hacks

zeha:


etc/zsh/zshenv
```diff
@@ -25,21 +25,6 @@
 # no xsource() here because it's only created in zshrc! (which is good)
 [[ -r /etc/environment ]] && source /etc/environment
 
-# make sure /usr/bin/id is available
-if [[ -x /usr/bin/id ]] ; then
-    [[ -z "$USER" ]]          && export USER=$(/usr/bin/id -un)
-    [[ $LOGNAME == LOGIN ]] && LOGNAME=$(/usr/bin/id -un)
-fi
```
jkirk, line 32: What problem was solved with this block? Why can we remove this now?

jkirk:
FTR, I created #180 for this.

etc/zsh/zshenv
zeha: it made sure USER was set, which i think was only necessary when started from file-rc.

i couldnt find cases where LOGNAME was not set.

mika: > What problem was solved with this block? Why can we remove this now?

I *think* that was a workaround for our old init situation on Grml back in the days(tm).
I _very_ much hope that this is no longer a problem for us nowadays

mika (approved):

PR #182, opened 2025-01-16
zsh: drop solaris support

zeha:


mika (approved):
We can easily revert this if we still have SunOS/Solaris users out there, so I'd say let's give this a try

PR #185, opened 2025-01-26
zshrc: remove "hbp" alias

zeha:
The target program (hg-buildpackage) does not exist anymore.


mika (approved):
(I like to use "zshrc:" in the beginning of the commit message to clarify what this is about)

Issue #183, opened 2025-01-24
Looks like `xunfunction` should be called at end of `zshrc`


xaocon:
It looks to me [like](https://github.com/grml/grml-etc-core/blob/1b0ef9cdefd28fe46ac38702e46d4da363375565/etc/zsh/zshrc#L492) `xunfunction` should be called near the end of `etc/zsh/zshrc`. I see [`unfunction grml_status_feature`](https://github.com/grml/grml-etc-core/blob/1b0ef9cdefd28fe46ac38702e46d4da363375565/etc/zsh/zshrc#L3648) so maybe it should be called there but `grml_status_feature` should be added as a function to remove?

zeha:
`xunfunction` should go away, is what I'm inferring from the git history.

mika:
> `xunfunction` should go away, is what I'm inferring from the git history.

@ft maybe you could help us out and see what we could do about the `xunfunction` function and how to deal with it? 😄

zeha:
https://github.com/grml/grml-etc-core/commit/724508d30c788da653579d44ac3566f27ffed1d6 tried to remove it, but didn't explain in detail why. But let's do it now.

mika:
@zeha from commit 724508d30c788da653579d44ac3566f27ffed1d6, @ft noted:

```
  * zshrc: remove xunfunction() altogether; it prevented things like
           'source ~/.zshrc' from working.
 ```

PR #186, opened 2025-01-27
zshrc: deprecate xunfunction

zeha:
It was introduced in commit 3d6452fdfc44310b5615eaf61c95cab4c27d9bf8 in 2007, and 724508d30c788da653579d44ac3566f27ffed1d6 in 2008 tried to remove it. The remove was reverted in 347c7def8583733dd8f8440214af25fbaf89666f in 2008.

Let's try this again, this time with a deprecation warning.

Addresses #183.

### DIFF
--- a/doc/grmlzshrc.t2t
+++ b/doc/grmlzshrc.t2t
@@ -858,6 +858,7 @@ arguments for details.
 
 : **xunfunction()**
 Removes the functions salias, xcat, xsource, xunfunction and zrcautoload.
+**Deprecated.**
 
 : **zrcautoload()**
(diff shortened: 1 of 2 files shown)


zeha:
@ft please also take a look

mika (commented):
LGTM, though would like feedback from @ft :)

ft (approved):
You are more careful than I would have been, with a deprecation period and all. :)

But it's fair all the same. I think we should make a point in removing this in like N months.

zeha:
Opened #187 so we don't forget to remove the function later.

PR #189, opened 2025-01-31
Switch manpage source to asciidoc

zeha:
Converted using txt2tags and additional manual cleanup.

Lets us reuse the asciidoc integration of the website builder.


mika (approved):
Good idea, thanks!

PR #190, opened 2025-02-01
CI: add debian, refcard jobs

zeha:
Minimal setup.

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -1,0 +1,33 @@
+name: test-build
+on:
+  pull_request:
+  push:
+
+concurrency:
+  group: "${{ github.ref }}"
+  cancel-in-progress: true
+jobs:
+  build-debian:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: "Update changelog"
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/test-build.yml
```diff
@@ -0,0 +1,58 @@
...
+
+  zsh-refcard:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install texlive
+        run: |
+          set -ex
+          sudo apt update
+          sudo apt install -y texlive-fonts-recommended texlive-latex-recommended
```
mika, line 43: apt always complains about "not meant for scripting usage" or alike, do we still want to use it?

mika (commented):
LGTM, only one question to clarify our handling of apt vs apt-get

.github/workflows/test-build.yml
zeha: I think this warning is gone in trixie and the interface didn't change.

mika: > I think this warning is gone in trixie and the interface didn't change.

Oh, wasn't aware of this! TIL, thx :)

mika (approved):

PR #191, opened 2025-02-08
aliases: apt-file list / ts $TIMEFMT

crpb:
afs alone is irritating
TS is convenient

### DIFF
--- a/etc/skel/.zshrc
+++ b/etc/skel/.zshrc
@@ -194,6 +194,7 @@ fi
 #alias -g SL='| sort | less'
 #alias -g S='| sort'
 #alias -g T='|tail'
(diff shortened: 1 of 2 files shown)


mika (changes requested):
Thanks, sounds useful.

No need to modify `debian/changelog`, we take care of this during release time, so please drop that commit :)

And while at it, we try to provide the according information in commit message (similar to what you provided with `debian/changelog` already), so `gbp dch` can use it as-is. So something like:

replace `alias for apt-file list` -> with `zshrc: alias afl for apt-file list`

replace `alias: global alias for timestamp` -> with `etc/skel/.zshrc: global alias TZ for timestamp/ts(1)`

mika (approved):
Nice, thanks!

PR #192, opened 2025-02-24
zshrc: Add alias llblk for lsblk with useful columns

jkirk:
TBH: I am not sure about the `#a2#` tag in the comment above. I just copied the comment from above.

The comment for `#a#` is not perfectly clear for me:
https://github.com/grml/grml-etc-core/blob/e2a5cafd7db6c24c72cac370bb073850b0a762f7/etc/zsh/zshrc#L42-L45


mika (approved):
LGTM, thx! (Would like to also have @zeha's +2, since that's what we're using from inside VMs to map disks from inside VMs  to the VM/host disks)

PR #196, opened 2025-03-28
tmux.conf: use brackets for if-shell commands

crpb:
* it's easier to read \ó/

### DIFF
--- a/etc/tmux.conf
+++ b/etc/tmux.conf
@@ -27,11 +27,17 @@ bind-key j join-pane -s !
 bind-key J join-pane
 
 ### Move current window to session named "bg" (and create session if it does not exist)
-bind-key B if-shell "! tmux has-session -t bg" "new-session -d -s bg" \; move-window -t bg
+bind-key B if-shell '! tmux has-session -t bg' {
+  new-session -d -s bg
+  move-window -t bg
+}
 
-### Reload Config
-if-shell "! (env | grep -q TMUX=/tmp/tmate)" \
-  "bind-key R source-file ~/.tmux.conf \\; source-file -q ~/.tmux.conf.local \\; display-message '~/.tmux.conf[.local] reloaded'"
+### Reload User Configs
+if-shell '! (env | grep -q TMUX=/tmp/tmate)' {
+  bind-key R source-file ~/.tmux.conf
+  source-file -q ~/.tmux.conf.local
+  display-message "~/.tmux.conf[.local] reloaded"
+}
 
 ###rebind keys
 bind-key h next-layout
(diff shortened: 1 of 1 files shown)


mika (commented):
Nice idea, thanks! Not sure how backwards compatible this is (you've confirmed with bullseye according to IRC), basically LGTM, leaving the +2 to @jkirk :)

mika:
From IRC: we seem to have issues with tmate, to be clarified :-/

crpb:
Sadly it seems (fooding old) tmate isn't liking any of those `{}` 
```
test@obelix:~$ grep '[{}]' /etc/tmux.conf
bind-key B if-shell '! tmux has-session -t bg' {
...
}
test@obelix:~$ tmate
lost server
```

ant it already complains about this new thing
https://github.com/grml/grml-etc-core/blob/d891ce58faf68905ec75dc9c409b51e139e30212/etc/tmux.conf#L65

```
cb@obelix ~dls % schroot --directory=/ --user root --chroot=source:app_bookworm
obelix# wget -O /etc/tmux.conf https://raw.githubusercontent.com/grml/grml-etc-core/refs/heads/master/etc/tmux.conf
--2025-03-28 15:46:34--  https://raw.githubusercontent.com/grml/grml-etc-core/refs/heads/master/etc/tmux.conf
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 2606:50c0:8003::154, 2606:50c0:8000::154, 2606:50c0:8001::154, ...
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|2606:50c0:8003::154|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 2844 (2.8K) [text/plain]
Saving to: '/etc/tmux.conf'

/etc/tmux.conf                              100%[========================================================================================>]   2.78K  --.-KB/s    in 0.003s

2025-03-28 15:46:34 (825 KB/s) - '/etc/tmux.conf' saved [2844/2844]

obelix# apt update; apt install -y tmate
....
obelix# tmate
/etc/tmux.conf:65: usage: if-shell [-bF] [-t target-pane] shell-command command [command]                                                                               [0/0]
Tip: if you wish to use tmate only for remote access, run: tmate -F
....
```

Maybe tmate needs to unlearn about /etc/tmux.conf 🙈 

https://salsa.debian.org/debian/tmate/-/blob/upstream/tmux.h?ref_type=heads#L56

crpb:
> also the tmate issue needs to be sorted

thougts? https://salsa.debian.org/cb/tmate/-/commit/0dae773b26e929bb448f29a60aeb8fe0b45c42a0

zeha:
did that land in forky?

i'd be ok with it being broken on trixie (if it works in forky), duno about the others.

zeha:
> i'd be ok with it being broken on trixie (if it works in forky), duno about the others.

on second thought, doing this for grml-etc-core would be bad. users expect better backwards compat.

crpb:
> did that land in forky?

no, i only created that here for me today to see if it works and it does.
but yeah, it would break even trixie if send in and then even accepted with this `/etc/tmux.con`

zeha (commented):
Needs to wait for Debian oldstable

zeha:
@mika wdyt about adding a `tmate` alias that expands to `tmate -f $HOME/.tmate.conf`, which could then be a custom tmate.conf?
Don't know if this is feasible, but maybe worth considering?

zeha:
> > @mika wdyt about adding a `tmate` alias that expands to `tmate -f $HOME/.tmate.conf`, which could then be a custom tmate.conf? Don't know if this is feasible, but maybe worth considering?
> 
> I love this idea and just implemented it. But tmate is _really stupid:_ it still reads the global `/etc/tmux.conf`, even when explictly running `tmate -f ...` :(

Ugh, that seems bad.

> So I don't see any way forward to continue usage of tmate with tmux :( Related: #247

So ... out with tmate?

zeha (approved):

zeha:
tmate support was dropped, so this should be able to land

mika (approved):

PR #197, opened 2025-04-04
zshrc: cleanup

crpb:
* fix output of $1 in H-Glob help
  * only trans() if /bin/trans isn't present

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -2625,8 +2625,10 @@ if [[ -r /etc/debian_version ]] ; then
           #a3# Execute \kbd{aptitude update ; aptitude safe-upgrade}
           salias -a up="aptitude update ; aptitude safe-upgrade"
         fi
-        #a3# Execute \kbd{dpkg-buildpackage}
-        alias dbp='dpkg-buildpackage'
+        if check_com -c dpkg-buildpackage ; then
+          #a3# Execute \kbd{dpkg-buildpackage}
+          alias dbp='dpkg-buildpackage'
+        fi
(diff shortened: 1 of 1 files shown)


etc/zsh/zshrc
```diff
@@ -2669,7 +2671,9 @@ fi
 # grmlstuff
 function grmlstuff () {
     #a1# Output version of running grml
-    alias grml-version='cat /etc/grml_version'
+    if test -f /etc/grml_version; then
```
mika, line 2674: that should be:

```
  if [[ -r /etc/grml_version ]] ; then
```

... to make it consistent with other code in this file  :)

etc/zsh/zshrc
```diff
@@ -3307,6 +3311,7 @@ fi
 #  $ awk -F ':' '{ print $2" : "$1" "$3 }' \
 #    /usr/local/lib/words/en-de.ISO-8859-1.vok > ~/.translate/de-en.ISO-8859-1.vok
 #f5# Translates a word
+if ! (( ${+commands[trans]} )); then
```
mika, line 3314: FTR: that seems to address `/usr/bin/trans` from the `translate-shell` package, sounds like a good idea.

Though maybe we should indent the lines from 3315 til 3329 then though?

mika (commented):
Thanks, good ideas, some nitpicks though. :)

The commit message is a bit misleading with `zshrc: check_com's for aliases` in its summary line, given that the change also includes several other changes :)

Would make sense to separate the changes by topic and make separate commits then, can of course all be part of one single PR :)

mika (approved):
Great, thanks!

PR #202, opened 2025-05-27
zshrc: allow disabling macOS `path_helper`

rvolosatovs:
After recently updating `grml` zsh config (among other things) on macOS it took me a while to figure out why tools on my `PATH` were acting strange.
Turned out the reason is #175 - `path_helper` is now called in `zshrc`, which breaks setups modifying `PATH` in `zprofile` or earlier (in my case https://github.com/nix-darwin/nix-darwin and `nix shell` functionality), because `path_helper` *prepends* its paths in `PATH`, making it impossible to e.g. use other `git` executable than provided by the OS without modifying `PATH` *after* loading `grml` zshrc.

Personally, I think `zprofile` is a better fit for `path_helper`, but regardless of where one adds this logic, users should be able to disable it (e.g. if already done externally), which is what I'm adding in this PR.

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -350,6 +350,7 @@ setopt unset
 NOCOR=${NOCOR:-0}
 NOETCHOSTS=${NOETCHOSTS:-0}
 NOMENU=${NOMENU:-0}
+NOPATHHELPER=${NOPATHHELPER:-0}
 NOPRECMD=${NOPRECMD:-0}
 COMMAND_NOT_FOUND=${COMMAND_NOT_FOUND:-0}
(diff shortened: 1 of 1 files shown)


mika:
> Does that mean that effectively grml does not support zshenv and zprofile and
> instead users are expected to reimplement that functionality in zshrc.local/zlogin?

IMO not really (we of course try to support those), but sometimes it's easier to handle customization in your `.zshrc.pre` and/or `.zshrc.local` :)

Having the option to skip usage of `/usr/libexec/path_helper` by setting `NOPATHHELPER=1` sounds useful and non-intrusive to me, so IMO this PR LGTM, any objections @zeha or @ft, otherwise I'd merge it as-is. :)

mika (approved):

zeha (approved):

PR #203, opened 2025-06-05
zshrc:cdt() use optional template directory

crpb:
for those who are left with 20+ `/tmp/tmp.XXXXX` dirs after an hour and don't wanna inspect each and every one or scan through `ls -l /tmp/tmp*` returns.


crpb:
narf....

ft:
Yeah, lgtm.

mika:
Thanks @ft for the review! :bow:

PR #205, opened 2025-06-07
zshrc: fix screen alias

crpb:


etc/zsh/zshrc
```diff
@@ -2449,8 +2449,8 @@ if check_com -c screen ; then
             alias screen='screen -c /etc/grml/screenrc'
         fi
     elif [[ ! -r $HOME/.screenrc ]] ; then
-        if [[ -r /etc/grml/screenrc_grml ]]; then
-            alias screen='screen -c /etc/grml/screenrc_grml'
+        if [[ -r /etc/grml/screenrc ]]; then
+            alias screen='screen -c /etc/grml/screenrc'
```
zeha, line 2453: this is exactly the same as below. why?

PR #207, opened 2025-06-19
zshrc: export all locale variables

zeha:
Previously only LANG LC_ALL LC_MESSAGES were exported. I assume the historic reason is that grml-setlang used to set up exactly these three variables in /etc/default/locale. With the new grml-setlang changes LC_COLLATE LC_TIME can also be set. Lets support all variables so this is not specific to grml-setlang.


mika (approved):
Great idea, thanks!

PR #213, opened 2025-06-19
Move grml-tips completion to grml-tips package

zeha:
grml-tips.deb should really install the completion for itself. Should make it easier to find / update, too.


debian/control
```diff
@@ -15,6 +15,8 @@ Bugs: mailto:bugs@grml.org
 
 Package: grml-etc-core
 Architecture: all
+Breaks:
+ grml-tips (<< 0.8.4)
```
mika, line 19: 0.8.4 is already released, shouldn't this then be << 0.8.5? :thinking:

zeha: Indeed, thanks for catching this

zeha: Sorry, had to merge this to avoid failing daily and ci builds due to the newly introduced break relation.

PR #223, opened 2025-07-26
Fix chpwd error when DIRSTACKSIZE is unset by Claude Code

MSch:
Add null check for DIRSTACKSIZE variable before numeric comparison to prevent shell errors when Claude Code unsets the environment variable.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -1537,6 +1537,7 @@ if zstyle -T ':grml:chpwd:dirstack' enable; then
 
     function chpwd () {
         (( ZSH_SUBSHELL )) && return
+        [[ -z $DIRSTACKSIZE ]] && return
         (( $DIRSTACKSIZE <= 0 )) && return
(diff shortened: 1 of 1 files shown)


msch-nutrient:
@zeha I wish I knew. I've been observing ``chpwd:2: bad math expression: operand expected at `<= 0 '`` from time to time when it's invoking the `Bash` tool (which actually runs `$SHELL`) and haven't been able to track down why exactly it happens. You could argue, that in the age of vibe coding it's easier to add more slop / defensive coding everywhere to make up for whatever the AI may or may not do.

Zenexer:
Glancing at the line that was erroring, I'm pretty sure `(( $DIRSTACKSIZE <= 0 ))` has a typo: that $ shouldn't be there.

Opened https://github.com/grml/grml-etc-core/pull/254 to fix the typo and remove the test introduced in #223, as it's not longer necessary once the typo is fixed.

PR #224, opened 2025-09-17
zsrhc hash directories only if they exist; add ~systemd

crpb:
* Why? On *BSD you won't have any need for ~linux, ~deb, ~systemd nor
   will ~slog or ~www exist on every system.

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -2408,13 +2408,14 @@ fi
 
 # 'hash' some often used directories
 #d# start
-hash -d deb=/var/cache/apt/archives
-hash -d doc=/usr/share/doc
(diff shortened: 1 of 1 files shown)


crpb:
> Is there any problem I'm missing? 🤔

install nginx/lighttpd/apache/whatever and the directoy will magically be created i head :-)

crpb:
> With this applied, however, does the reference card still generate correctly?

this lead me to #225  🤣  where i was the culprit 

but the hashes look good here with this pr checked out and running `make test`.

zeha:
I don't have an opinion on which directories are added here really. I'd question some of the existing ones as maybe obsolete?

I am mildly concerned with the amount of stuff the config `stat`s on startup.

PR #226, opened 2025-09-22
Remove unused code from script-functions

zeha:
I imagine when we move the remaining scripts from grml-scripts into grml-live, `script-functions` should be merged with `autoconfig.functions`. Until then we can at least slim this file down.

### DIFF
--- a/etc/grml/script-functions
+++ b/etc/grml/script-functions
@@ -103,16 +103,6 @@ checkbootparam(){
 }
 # }}}
 
-# {{{ check whether $1 is yes
-checkvalue(){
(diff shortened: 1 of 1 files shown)


ft:
I didn't check, but if any of these are in the manual page for grml's zshrc, they should be removed also.

zeha:
> I didn't check, but if any of these are in the manual page for grml's zshrc, they should be removed also.

the zshrc config does not source script-functions. I expect them to not be in the man page, or if they are then they are already duplicated in zshrc (and will stay in there).

zeha:
Yeah, it's one of the long standing issue I'd like to clean up :)

mika (approved):

PR #229, opened 2025-10-04
zshrc: asc() check for existance of autossh

crpb:


zeha (changes requested):
still shouldn't override an existing `asc` command, i imaginr

zeha (approved):

crpb:
i never used it aswell. i just stumbled over a function which couldn't even work as `autossh` isn't installed. i only sometimes use ssh-ping..

Issue #230, opened 2025-10-05
zshrc: asc?


zeha:
As noted in https://github.com/grml/grml-etc-core/pull/229 maybe `asc` has become obsolete.

@ft @mika wdyt?

PR #231, opened 2025-10-06
Remove "asc" function

zeha:
Used GNU screen. Probably users switched to tmux and/or mosh.

Fixes https://github.com/grml/grml-etc-core/issues/230


mika (approved):
(FTR: I like to have "zshrc: " at the beginning of  the commit message summary, so we always know what this is about, also makes writing release notes a bit easier then :))

PR #245, opened 2026-02-15
change formatting of final 'if-shell' command to be compatible with tpm

### LINKED ISSUE #244: Tmux Plugin Manager (tpm) will no longer work from within .tmux.conf.local
I only recently upgraded to the latest version of [.tmux.conf](https://grml.org/console/tmux.conf), and found that the last two lines in PR #152 break the [Tmux Plugin Manager](https://github.com/tmux-plugins/tpm) (tpm) when it's loaded from within ~/.tmux.conf.local. The result is that this PR requires one to either move their tpm plugin configuration to .tmux.conf (which defeats the purpose of isolating user modications to .tmux.conf.local), or to change: 
```tmux
if-shell "! (env | grep -q TMUX=/tmp/tmate)" \
  "source-file -q ~/.tmux.conf.local"
```
back to:
```tmux
source-file -q ~/.tmux.conf.local
```
Any ideas on how to support both `tmate` and `tpm`?

tomhoover:
After the changes introduced in PR #152, [Tmux Plugin Manager](https://github.com/tmux-plugins/tpm) will no longer work when loaded from within ~/.tmux.conf.local. The result is that #152 requires one to either move their tpm plugin configuration to .tmux.conf (which defeats the purpose of isolating user modications to .tmux.conf.local), or to change the formatting of the final 'if-shell' command.

Closes #244

### DIFF
--- a/etc/tmux.conf
+++ b/etc/tmux.conf
@@ -71,5 +71,6 @@ set-window-option -g window-status-style fg=blue,bg=black
 set-window-option -g window-status-current-style bold
 
 ### source user-specific local configuration file
-if-shell "! (env | grep -q TMUX=/tmp/tmate)" \
-  "source-file -q ~/.tmux.conf.local"
+if-shell "! (env | grep -q TMUX=/tmp/tmate)" {
+  source-file -q ~/.tmux.conf.local
+}


tomhoover:
> We should verify that we don't break tmate (most notably als /cc @crpb @jkirk who are aware of the tmux/tmate situation also :))
> 
> (I'm really not happy that tmate is based on an old tmux version and still always reads in the main tmux configuration file, I wasn't aware of tpm at all, thanks for also bringing this to our attention!)

Before creating #244 and this PR, I had only searched for issues/PRs mentioning `tpm`. Since none mentioned it, I continued with the creation this PR. I've now found the discussion at https://github.com/grml/grml-etc-core/pull/196#issuecomment-2761777079 that indicates tmate doesn't like `{}`, so this PR may not be something you'd want to merge.

Issue #251, opened 2026-04-19
grml-screenrc needs changes for screen 5.0


akorn:
GNU Screen 5.x introduced several incompatible changes: for example, colours have to be specified numerically now, and the `%l` string escape to get the load average is gone.

I got started on converting the old screenrc and it's now usable for me, but still not quite the same as it used to be. (In particular, the "bell in window X" message is still messed up.)

I'm sharing what I have so others can use it as a starting point, but it's unlikely I'll be able to commit more of my own time to refining it.

Ideally we'd also be able to support screen 4.x and 5.0 concurrently...

```
# while this works, forking to run `/bin/cut` once every 5s feels expensive and wasteful
backtick 3 5 5  /bin/cut -d' ' -f1-3 /proc/loadavg
caption always "%{+b}%{= 9;0}$USER@%{= 15;0}%H | %{= 11;0}(load: %3` | %{= 9;0}cpu: %1` | %{= 10;0}net: %2`)  %-21=%{= 15;0}%D %Y-%m-%d %0c"
hardstatus alwayslastline "%{= 1;7 }%n%f %t %{= 7;0} | %?%-Lw%?%{= 4;7}%n%f %t%?(%u)%?%{= 7;0}%?%+Lw%? %{= 7;0}"
```

mika:
Related, see https://bugs.debian.org/1082656 + https://tracker.debian.org/news/1743013/accepted-screen-501-1-source-into-unstable/ and the upstream announce of screen 5.0.0  at https://www.mail-archive.com/info-gnu@gnu.org/msg03317.html

Regarding the load average removal I found https://www.mail-archive.com/screen-users@gnu.org/msg04670.html + https://savannah.gnu.org/bugs/?66003 -> https://cgit.git.savannah.gnu.org/cgit/screen.git/commit/?id=99a8b26d7ae221dfdc1bbe8f04f5867f7ebc07bc

This is all rather unfortunate and I'm wondering whether we can actually support screen versions before 5 and >=5 concurrently :-/

akorn:
A wrapper function/alias could decide which screenrc to load depending on the version of screen installed, but I'm not sure it's worth it.

I suppose the grml-testing release will have to ship a screenrc for screen5, while stable stays with screen4 until the 5.0 screen package hits Debian stable. (I realize this is also not ideal.)

mika:
[...]

PR #252, opened 2026-06-01
Support screen 5.0

zeha:
Add loadavg-screen backtick program, as %l is gone. The colors/formatting of the caption/hardstatus lines are almost the same, barring brightness on green. Some things like +b seem to no longer work as documented, so put b into each color string. % no longer seems to need the -21 (and actually break when thats present).

Additionally, screen v5 seems to put "*" into the %f flags of the currently selected window, so we don't have to do it ourselves anymore. OTOH we get a useless "*" in the flags of the leftmost current window display too.

Maybe we don't need screenrc_v4 at all?

### DIFF
--- a/debian/grml-etc-core.links
+++ b/debian/grml-etc-core.links
@@ -1,2 +1,3 @@
 /usr/share/man/man1/grml-scripts-core.1.gz /usr/share/man/man1/cpu-screen.1.gz
 /usr/share/man/man1/grml-scripts-core.1.gz /usr/share/man/man1/ip-screen.1.gz
+/usr/share/man/man1/grml-scripts-core.1.gz /usr/share/man/man1/loadavg-screen.1.gz
--- a/etc/grml/screenrc
+++ b/etc/grml/screenrc
@@ -38,11 +38,6 @@
 
 # shell                 /bin/zsh
 
-# "sorendition": set the colors for
-# the "messages" and "text marking"
-# (ie text you mark in copy mode):
-  sorendition 10 99 # default!
-
 # use %n to display the window number and %t for its title:
   activity              "activity in %n (%t) [%w:%s]~"
 
--- a/etc/grml/screenrc
+++ b/etc/grml/screenrc
@@ -21,7 +21,6 @@
 # defsilence            off             # default: off
 # hardcopy_append       on              # default: off
   hardcopy_append       off             # default: off
-  nethack               on              # default: off
 # don't display the copyright page
   startup_message       off             # default: on
 # no annoying audible bell, please
--- a/etc/grml/screenrc
+++ b/etc/grml/screenrc
@@ -5,14 +5,15 @@
 # License:       This file is licensed under the GPL v2.
 ################################################################################
 
-  backtick 1 0 60   /usr/bin/cpu-screen
-  backtick 2 0 60   /usr/bin/ip-screen
-  caption always "%{+b rk}$USER@%{wk}%H | %{yk}(load: %l |%{rk} cpu: %1` | %{Gk}net: %2`)  %-21=%{wk}%D %Y-%m-%d %0c"
-  hardstatus alwayslastline "%{wr}%n%f %t %{kw} | %?%-Lw%?%{wb}%n*%f %t%?(%u)%?%{kw}%?%+Lw%? %{wk}"
+  backtick 1 0 60   /usr/bin/loadavg-screen
+  backtick 2 0 60   /usr/bin/cpu-screen
+  backtick 3 0 60   /usr/bin/ip-screen
+  caption always "%{b1;0}$USER@%{b7;0}%H | %{b3;0}(load:%1` |%{b1;0} cpu: %2` | %{b2;0}net: %3`) %=%{b7;0}%D %Y-%m-%d %0c "
+  hardstatus alwayslastline "%{1;7}%n%f %t %{7;0} | %-Lw%{4;7}%n%f %t%?(%u)%?%{7;0}%+Lw%="
 
 # switch order of caption and hardstatus:
-#  hardstatus alwayslastline "%{+b kr}$USER@%{kw}%H | %{ky}(load: %l |%{kr} cpu: %1` | %{kG}net: %2`)  %-21=%{kw}%D %Y-%m-%d %0c"
-#  caption always "%{rw}%n%f %t %{wk} | %?%-Lw%?%{bw}%n*%f %t%?(%u)%?%{wk}%?%+Lw%? %{wk}"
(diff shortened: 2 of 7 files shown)


zeha:
openqa-tests will need new needles after this for the flags change :-/

mika:
Uh nice, thx! Would be great if we could even avoid usage/support of `screenrc_v4`!

So far I only noticed two issues (unless I mis-configured anything when taking over your changes into my local daily Grml ISO as of today):

1) It looks like it highlights the current window (`1`) until the end (`2`), instead of just the current window (`1`)?

screenshot 2026-06-01T17-57-27

2) The screenrc we ship via grml-live's `config/files/GRMLBASE/etc/skel/.screenrc` seems to fail yet (so it might also need an update with caption/hardstatus, but maybe there's even a way how to get rid of this file?):

screenshot 2026-06-01T17-57-18

zeha:
> Would be great if we could even avoid usage/support of `screenrc_v4`!

I think the only question is if users want screenrc and/or screenrc_generic? Without the /usr/bin scripts screenrc (and thus screenrc_v4) is probably pointless?

> So far I only noticed two issues (unless I mis-configured anything when taking over your changes into my local daily Grml ISO as of today):
> 
> 1. It looks like it highlights the current window (`1`) until the end (`2`), instead of just the current window (`1`)?

Indeed. fixed now.

> 2. The screenrc we ship via grml-live's `config/files/GRMLBASE/etc/skel/.screenrc` seems to fail yet (so it might also need an update with caption/hardstatus, but maybe there's even a way how to get rid of this file?):

Yeah that needs an update. I guess we have this file "only" to present a blue username instead of a red user? maybe we can drop that difference?

zeha:
Should close #251 when merging.

jkirk:
> > Would be great if we could even avoid usage/support of `screenrc_v4`!
> 
> I think the only question is if users want screenrc and/or screenrc_generic? Without the /usr/bin scripts screenrc (and thus screenrc_v4) is probably pointless?

I haven't used GNU/screen for quite some time now. Is there a logic where `screenrc_generic_v4`  and `screenrc_v4` is selected when screen 4.x is used or am I supposed to do it manually?

> > So far I only noticed two issues (unless I mis-configured anything when taking over your changes into my local daily Grml ISO as of today):
> > 
> > 1. It looks like it highlights the current window (`1`) until the end (`2`), instead of just the current window (`1`)?
> 
> Indeed. fixed now.

Not sure, if I am holding it wrong, but this is how it looks like if I apply the setting to screen (5.0.1) in Grml daily:

Screenshot from 2026-06-07 23-42-18

But to be fair, screen looks "funny" with the default settings (in the Grml daily used):

Screenshot from 2026-06-08 00-01-41

Is this how it should look like?

> > 2. The screenrc we ship via grml-live's `config/files/GRMLBASE/etc/skel/.screenrc` seems to fail yet (so it might also need an update with caption/hardstatus, but maybe there's even a way how to get rid of this file?):
> 
> Yeah that needs an update. I guess we have this file "only" to present a blue username instead of a red user? maybe we can drop that difference?

Again, I am not a screen-user anymore, but I think this was nice feature.

zeha:
> > > Would be great if we could even avoid usage/support of `screenrc_v4`!
> > 
> > 
> > I think the only question is if users want screenrc and/or screenrc_generic? Without the /usr/bin scripts screenrc (and thus screenrc_v4) is probably pointless?
> 
> I haven't used GNU/screen for quite some time now. Is there a logic where `screenrc_generic_v4` and `screenrc_v4` is selected when screen 4.x is used or am I supposed to do it manually?

For now: Manually. There is no way in the screen config of dealing with this.

Maybe zshrc could detect and alias it, but that seems like a lot of magic.

> > > So far I only noticed two issues (unless I mis-configured anything when taking over your changes into my local daily Grml ISO as of today):
> > > 
> > > 1. It looks like it highlights the current window (`1`) until the end (`2`), instead of just the current window (`1`)?
> > 
> > 
> > Indeed. fixed now.
> 
> Not sure, if I am holding it wrong, but this is how it looks like if I apply the setting to screen (5.0.1) in Grml daily:
> 
> Screenshot from 2026-06-07 23-42-18

On ISOs, you also have to fixup the screenrc installed by grml-live.

> But to be fair, screen looks "funny" with the default settings (in the Grml daily used):
> 
> Screenshot from 2026-06-08 00-01-41
> Is this how it should look like?

This is the effect of using the (existing) v4 config on v5. (Which is why we are here in the first place.)

crpb:
fwiw, just installed a build of this mr and it's looking good so far.
maim_20260613-233420
(those two screens are actually two panes in my tmux-session ^_^)

mika (approved):
Thanks, LGTM for root user, verified on Debian/testing based daily Grml ISO.

We shouldn't forget about our `/etc/skel./.screenrc` we ship via grml-live nowadays, as noted in https://github.com/grml/grml-etc-core/pull/252#issuecomment-4594314738

zeha:
> We shouldn't forget about our `/etc/skel./.screenrc` we ship via grml-live nowadays, as noted in [#252 (comment)](https://github.com/grml/grml-etc-core/pull/252#issuecomment-4594314738)

I was wondering if we can do a bit of sed magic in d/rules to produce the user screenrc. grml-live can still copy it into the right place, but then we don't have two places with the almost same file. (I know I probably moved the file incl. the rest of skel to grml-live...)

zeha:
@mika could you update your console blog post?

mika:
> @mika could you update your console blog post?

@zeha @jkirk Sure, thanks for the reminder! So we currently have `wget -O .screenrc   https://grml.org/console/screenrc` which ends up as redirect to `https://git.grml.org/f/grml-etc-core/etc/grml/screenrc_generic`.

We're handling this through https://github.com/grml/grml.org/blob/master/static/console/.htaccess, and also have https://github.com/grml/grml.org/blob/master/static/scripts/grml-config.sh in place.

Should we add support for `wget -O .screenrc   https://grml.org/console/screenrc_v4`  for the ones running older GNU screen versions, or is there any better approach we could come up with? :thinking:

zeha:
> Should we add support for `wget -O .screenrc https://grml.org/console/screenrc_v4` for the ones running older GNU screen versions, or is there any better approach we could come up with? 🤔

I think that's the best we can do :(

zeha:
> > Should we add support for `wget -O .screenrc https://grml.org/console/screenrc_v4` for the ones running older GNU screen versions, or is there any better approach we could come up with? 🤔
> 
> I think that's the best we can do :(

https://github.com/grml/grml.org/pull/154 should add the redirect

PR #254, opened 2026-06-02
zshrc: dirstack: fix chpwd math expression for unset DIRSTACKSIZE

### LINKED ISSUE #253: Dollar sign in (( )) causes error if DIRSTACKSIZE is unset
https://github.com/grml/grml-etc-core/blob/48f12102d51fe38e28439dda0160c060029a6aa9/etc/zsh/zshrc#L1544

It looks like this should be `(( DIRSTACKSIZE <= 0 ))`, not `(( $DIRSTACKSIZE <= 0 ))`. The current version can cause weird errors such as this if DIRSTACKSIZE is ever unset:

```
chpwd:2: bad math expression: operand expected at `<= 0 '
```

Zenexer:
Fixes #253.

`(( $DIRSTACKSIZE <= 0 ))` in the `chpwd` hook does literal text substitution before the arithmetic parser runs. When `DIRSTACKSIZE` is unset or empty, the expression becomes `(( <= 0 ))` and zsh errors with:

```
chpwd:2: bad math expression: operand expected at `<= 0 '
```

Also removes the `[[ -z $DIRSTACKSIZE ]] && return` line added in https://github.com/grml/grml-etc-core/pull/223: it looks like that was an AI-made attempt at fixing the same bug, but it's a workaround rather than a proper fix--the $ in the (( )) conditional shouldn't be there.

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -1540,8 +1540,7 @@ if zstyle -T ':grml:chpwd:dirstack' enable; then
 
     function chpwd () {
         (( ZSH_SUBSHELL )) && return
-        [[ -z $DIRSTACKSIZE ]] && return
-        (( $DIRSTACKSIZE <= 0 )) && return
+        (( DIRSTACKSIZE <= 0 )) && return
         [[ -z $DIRSTACKFILE ]] && return
         grml_dirstack_filter $PWD && return
         GRML_PERSISTENT_DIRSTACK=(


zeha:
Changes done using AI/LLMs need to be submitted using Author:/Co-authored-by: tags set appropriately.

zeha:
LGTM apart from the missing LLM attribution.

mika:
Pulling in also @ft (TBH, I don't really understand what's going on here)

zeha:
> Pulling in also @ft (TBH, I don't really understand what's going on here)

If we assume that users (or their LLM agents) unset DIRSTACKSIZE, then I guess this is a valid problem. But I dunno why they would do this.

mika:
> > Pulling in also @ft (TBH, I don't really understand what's going on here)
> 
> If we assume that users (or their LLM agents) unset DIRSTACKSIZE, then I guess this is a valid problem. But I dunno why they would do this.

Ahhhh, now that I looked as this with `--word-diff`, I finally also see the `(( [-$DIRSTACKSIZE-]{+DIRSTACKSIZE+} <= 0 ))`, looked like a revert to me until then (guess it's time for a break :)), thanks!

MSch:
> If we assume that users (or their LLM agents) unset DIRSTACKSIZE, then I guess this is a valid problem. But I dunno why they would do this.

I submitted #223 originally - I never figured out what was going on, just that I got this error when Claude invoked it, so I added the minimum viable defensive slop.

I don't know what Claude does exactly to the shell env, but Codex does a bunch of work around snapshotting the zsh env that also bit me, see https://github.com/openai/codex/pull/16974/, so… maybe something at some point in time did indeed unset `DIRSTACKSIZE`.

zeha:
What I don't understand is why Codex/claude-code/... have to use the users shell in the first place. Why not start a clean /bin/sh or avoid shell altogether. m(

MSch:
@zeha from the docs https://developers.openai.com/codex/config-basic
> Snapshot your shell environment to speed up repeated commands

Why the inherited PATH isn't enough also isn't clear to me. I guess when the agent changes working dir they need to simulate cd in case tools like asdf/mise/etc change PATH?

Zenexer:
> Changes done using AI/LLMs need to be submitted using Author:/Co-authored-by: tags set appropriately.

I wrote the fix and only had the LLM open the PR--and then decided I didn't like its wording and rewrote the PR message anyway. Still want attribution?

> What I don't understand is why Codex/claude-code/... have to use the users shell in the first place. Why not start a clean /bin/sh or avoid shell altogether. m(

It just depends what's being done. It can be handy for, say, figuring out where the heck the grml repo is and opening a PR. ;)  It needs my ssh-agent socket to push the PR.

zeha (approved):
lgtm

ft:
Sorry for the late reply — I am unfortunately not checking in very frequently…

This change looks fair to me.

PR #255, opened 2026-07-24
zshrc: check for /etc/os-release instead of /etc/debian_version

mika:
Looks like `/etc/debian_version` might be disappearing from base-files, see https://lists.ubuntu.com/archives/ubuntu-devel/2026-July/043633.html

(Thanks @zeha for the pointer)

### DIFF
--- a/etc/zsh/zshrc
+++ b/etc/zsh/zshrc
@@ -2540,7 +2540,7 @@ Enjoy your grml system with the zsh!$reset_color"
 }
 
 # debian stuff
-if [[ -r /etc/debian_version ]] ; then
+if grep -q -e '^ID=debian$' -e '^ID_LIKE=.*debian' /etc/os-release 2>/dev/null ; then
     if [[ -z "$GRML_NO_APT_ALIASES" ]]; then
         #a3# Execute \kbd{apt-cache policy}
         alias acp='apt-cache policy'


zeha:
I imagine on Ubuntu this will fail, but whatever. They get what they sow...

zeha (approved):

mika:
> I imagine on Ubuntu this will fail, but whatever. They get what they sow...

Ah good point, on Ubuntu we seem to have:

```
ID=ubuntu
ID_LIKE=debian
```

Maybe we should also check for `ID_LIKE=debian`?

zeha:
> Maybe we should also check for `ID_LIKE=debian`?

🤷🏼

etc/zsh/zshrc
```diff
@@ -2540,7 +2540,7 @@ Enjoy your grml system with the zsh!$reset_color"
 }
 
 # debian stuff
-if [[ -r /etc/debian_version ]] ; then
+if grep -q '^ID.*=debian' /etc/os-release 2>/dev/null ; then
```
jkirk, line 2543: LGTM.

Didn't know about `ID_LIKE`, but make sense to make the config work on Ubuntu too.

Just curious: since when does `/etc/os_release` exists? (Which is the last release without one?)

jkirk (approved):

PR #256, opened 2026-08-02
loadavg-screen: fix which fields are output

akorn:
The previous version, `uptime | cut -f 4- -d:`, produced output like

`38,  1 user,  load average: 0.23, 0.35, 0.20`

where the "38" is the minute field from " 11:49:41 up 5 days,  4:38,  1 user,  load average: 0.23, 0.35, 0.20".

Since `screenrc` already causes `load:` to be included in the status line, there is also no need to include the `load average: ` output from `uptime`.

We only need the first three fields from /proc/loadavg and don't have to rely on the `uptime` program at all; this also saves us a fork(). (Which is why I'd in fact prefer to put `cut -d' ' -f1-3 /proc/loadavg` into screenrc instead of having screen call a shell script.)

### DIFF
--- a/usr_bin/loadavg-screen
+++ b/usr_bin/loadavg-screen
@@ -6,4 +6,4 @@
 # License:       This file is licensed under the GPL v2.
 #******************************************************************************
 
-cut -d' ' -f1-3 /proc/loadavg
+LC_ALL=C uptime | sed 's/.*: //'
--- a/usr_bin/loadavg-screen
+++ b/usr_bin/loadavg-screen
@@ -6,4 +6,4 @@
 # License:       This file is licensed under the GPL v2.
 #******************************************************************************
 
-uptime | cut -f 4- -d':' -
+cut -d' ' -f1-3 /proc/loadavg


zeha:
So the `uptime` call is not portable across OSes, but the `cut /proc/loadavg` also isn't. Not sure what will be portable

zeha:
I imagine this will work:

`LC_ALL=C uptime | cut -f 4- -d',' | cut -f 2- -d':'`

zeha:
Here are some `LC_ALL=C uptime` outputs:

- Debian 13: ` 12:23:13 up 89 days, 23:56,  5 users,  load average: 0.01, 0.03, 0.00`
- Debian 13 fresh boot: ` 12:24:40 up 0 min,  1 user,  load average: 0.37, 0.09, 0.03`
- macOS: `12:22  up 65 days, 17:07, 10 users, load averages: 3.47 7.12 13.55`
- FreeBSD 15.1: `10:22AM  up 3 mins, 2 users, load averages: 0.36, 0.50, 0.23`

zeha:
`LC_ALL=C uptime | sed 's/.*load averages*: *//'` is kinda meh but should work. WDYT?

akorn:
I was just typing that `LC_ALL=C uptime | sed 's/.*: //'` would produce correct output for all of the ones you mentioned.

I don't think we need to match "load averages*:" explicitly:

1. it's either the last field, in which case ".*:" is enough;
2. or it isn't, in which case we need something (slightly) more complex to get the actual loadavg out anyway.

zeha:
Sounds good to me

PR #259, opened 2026-08-08
zshrc: update alias screen

crpb:
* Remove obsolete references to screenrc_grml.
* Allow for a personal screenrc for root aswell.
* Use the v4 file if we need to.


mika (approved):
Good idea, thanks!

